PR #48, opened 2012-03-04
Set layout after adding each new pane

iNecas:
Especially when using tiled layout, creating more panes ends up with message:
"pane too small". For example:

```
tabs:
  - shells:
      layout: tiled
      panes:
        - echo 1
        - echo 2
        - echo 3
        - echo 4
        - echo 5
        - echo 6
```

It's because the layout is set after creating all the panes. Setting layout
after creating each pane fixes this issue.

### DIFF
--- a/lib/tmuxinator/assets/tmux_config.tmux
+++ b/lib/tmuxinator/assets/tmux_config.tmux
@@ -30,8 +30,8 @@ tmux <%= socket %> new-window -t <%= window(i+2) %> -n <%=s tab.name %>
 tmux <%= socket %> splitw -t <%= window(i+1) %>
 <%=      send_keys(tab.pre, i+1) if tab.pre %>
 <%=      send_keys(pane, i+1) %>
-<%     end %>
 tmux <%= socket %> select-layout -t <%= window(i+1) %> <%=s tab.layout %>
+<%     end %>
 <%   end %>
 <% end %>
 


primalmotion:
sorry to exhume this issue but this is still happening to me with tmuxinator 0.11.1 and tmux 2.7.

here is what I have

```yaml
name: apomux
tmux_options: -f .apomux.tmux.conf

windows:
    - core:
        layout: tiled
        panes:
          - squall:
            - clear
          - midgard:
            - clear
          - tidus:
            - clear
          - zack:
            - clear
          - wutai-internal:
            - clear
          - vince:
            - clear
          - barret:
            - clear
          - wutai:
            - clear
          - jenova:
            - clear
          - highwind:
            - clear
          - nanaki:
            - clear
          - yuffie:
            - clear
          - sephiroth:
            - clear
          - tagle:
            - clear
          - marlene:
            - clear
          - tifa:
            - clear
          - meister:
            - clear
          - apotests:
            - clear
          - apoctl:
            - clear
          - empty:
            - clear
```

This says this 
```
create pane failed: pane too small
can't find pane 18
can't find pane 18
can't find pane 18
create pane failed: pane too small
can't find pane 19
can't find pane 19
can't find pane 19
create pane failed: pane too small
can't find pane 20
can't find pane 20
can't find pane 20
```

and the window looks like this

screen shot 2018-04-18 at 16 44 52

This is on iTerm2 with tmux mode, but on Apple's Terminal it's the same problem (minus the weird gray bottom bar)

ethagnawl:
@primalmotion I'm pretty sure the culprit is `split-window`'s default behavior. 

From the manpage:

> split-window:
> Create a new pane by splitting target-pane: -h does a horizontal split and -v a vertical split; if neither is specified, **_-v is assumed_**.

As things stand, tmuxinator splits the pane vertically and then sets the layout. So, in your case it's running out of vertical space and the command is failing before it even tries to use the available space.

I just did a quick experiment which used `tmux splitw -h ...` to explicitly split the previous pane horizontally and your use case works -- as you're expecting it to. This can't just be wantonly implemented, though, because it will likely break other peoples' currently "working" layouts. I'll have to think on this and revisit.

mutlubicer:
I have 32 panes (8 (row) x 4 (columns)) in one of my tmux windows. All the panes are equal in size.
After upgrading to tmuxinator 0.8.1 (from whatever the version in Ubuntu 16.04), I started getting this error for almost the half of the panes in the window.

Then I generated the shell script from tmuxinator (tmuxinator debug <config-file>), and tried;
1. Saved the output to a file and ran it.
Result: I saw the same error.
2.a. Manually executed the initial commands that creates tmux session.
2.b. Attached to the session from another terminal.
2.c. Ran the rest commands.
Result: I didn't see any error. Panes were created successfully.
3. In the script, I copied the attach command to the line after session creation. So, the script creates session and attaches to it first. Then I manually detach from the session. The script continues and creates windows and panes, and then it attaches back.
Result: I didn't see any error. Panes were created successfully.

So, it looks like the panes are not correctly (re)sized when we don't attach. The resolution of terminal or the monitor actually doesn't matter in my case. Possibly the default size before attaching is too small?

ethagnawl:
Hi, @mutlubicer. I believe you're seeing the issue covered by #651. If you haven't already, please have a look at this issue and the suggested fix.

PR #81, opened 2012-10-14
Optionally persist per-shell history across tmux sessions

kmowery:
Adds a new configuration option "save_history", which enables storing per-shell history in root_dir.

This is accomplished by setting HISTFILE to special values in tmux_config.mux. When possible, filenames are taken from tab names. New shells created in the tmux session will use the standard history file (.bash_history by default), since I don't think tmuxinator can affect those shells.

I've tested on bash; it should work on zsh. Presumably, though, zsh users will be less interested, since it supports shared history far better than bash.

Input on style/correctness welcome.

### DIFF
--- a/lib/tmuxinator/assets/sample.yml
+++ b/lib/tmuxinator/assets/sample.yml
@@ -4,6 +4,7 @@
 project_name: Tmuxinator
 project_root: ~/code/rails_project
 socket_name: foo # Not needed.  Remove to use default socket
+save_history: true # Persist history for each pane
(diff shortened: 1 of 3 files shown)


Soliah:
If people still want this I'd appreciate it if someone could rebase this on top of current HEAD.

bedge:
sigh...just what I needed.
Better than manually setting HISTFILE before each shell.
I'll put the rebase on my todo list. (not holding breath)

Issue #31, opened 2011-11-04
load project from $cwd/.tmuxinator, then ~/.tmuxinator


hedgehog:
Expanding the project search path, to first inspect any `.tmuxinator` folder beneath the current directory, will allow a tmuxinator project to be included with the source code of the project they relate to.

PR #180, opened 2013-12-25
Return ability to load configs from local .tmuxinator dir.

svetlyak40wt:
This ability was introduced in #31 request and then disappeared during
code refactoring, I belive.


tarky:
I also want this feature.

PR #203, opened 2014-02-11
Support config file inside a project

mahemoff:
Wanting to ship a project's tmuxinator config with the project, so would like a way people can directly open the file as well as specify a project root relative to the config file. The new FAQ entry in the README explains how someone could use it.


nicolaiskogheim:
This would have been really, really nice.

PR #262, opened 2014-10-17
Add recent information about completion for zsh to README.md

ohcibi:
The script that is currently in the repository does not work for recent
zsh versions. The completion stuff is handled by another project. The
readme got updated to include this information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,17 @@ Add the following to your `~/.bashrc`:
 
 ### zsh
 
-Add the following to your `~/.zshrc`:
+The ZSH completion is maintained by the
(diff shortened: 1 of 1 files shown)


ohcibi:
I get
> undefined method `command_hash'
when I run tmuxinator in this version. Using ruby-2.1.5

ohcibi:
NVM. Had to uninstall 0.6.9 first.

ohcibi:
Ok. It works. I've added a feature in the zsh-completions repo which would print all projects for simply running `tmuxinator` in a projects section, which I will create a separate pull request for.

The new completions script should be pushed to zsh-completions anyways as some other repos and libraries (e.g. prezto) rely on it.

PR #273, opened 2014-12-15
add param to 'start' that starts project with custom session name/socket

### LINKED ISSUE #143: Start new session with same config as an existing project
I have a tmuxinator project called 'dev' which is like a skeleton dev setup I can use for lots of different purposes. When I call `mux start dev` it gets me into one session. I want something like `mux start-clone dev` that gets me into a new session with all the same tmuxinator config, only a different tmux session.

Is there a way to do this already?

jlipps:
this allows projects to be reused
fix #143


magicalbanana:
Has the new gem version been updated? And will there be a new field on the config yaml file? 

What I want to know is, is there's an option currently like when you would do:

```
tmux new-session -t session_name
```

So that even though it's the same session, you don't have the same windows open.

Soliah:
@magicalbanana I have cut and released 0.6.11.

`mux start project someothername` should work.

jlipps:
:+1:  thanks for the merge! Been using this the last 6 months and it's an integral part of my dev cycle. Now the only feature that I'm missing is passing params from the command line into argument slots inside the yaml config.

PR #279, opened 2015-01-06
Add fish autocompletion for mux

igas:
Fish only autocompletes if script has the same name as binary, we can't place autocompletion for both `mux` and `tmuxinator` in one file =(


rominf:
Hi, @igas.
It's ironic but I just solved this by myself (see #280). I think my solution is more verbose, but more generic and more DRY.

Issue #224, opened 2014-04-26
Autostart a Tmuxinator project on bootup under Ubuntu


jonasrk:
Hey there, thanks for your work.

I know this is not a tmuxinator bug, but can someone help me with this issue nevertheless? I did not get any feedback on stackoverflow ...

http://stackoverflow.com/questions/23269764/autostart-a-tmuxinator-project-on-bootup-under-ubuntu

Copy of the Stackoverflow text:

I have configured a __Tmuxinator project__ I want to have running on my server at all times. What bugs me is that I did not manage to get it to __start automatically__ after (unexpected) reboots. Here's what I have tried so far:

1. Editing __/etc/rc.local__ so that it now contains:

        mux start project &
        exit 0

2. (Following [this guide](http://www.stuffaboutcode.com/2012/06/raspberry-pi-run-program-at-start-up.html)) creating a script that contains
      
        case "$1" in
            start)
            echo "Starting muxproject"
            mux start project
            ;;
and configuring it with

        sudo update-rc.d NameOfYourScript defaults
3. (From the same guide:) Editing __etc/xdg/lxsession/LXDE/autostart__ and adding the line
    
        @mux start project
to it.

None of the above has worked. How can this be achieved?

Soliah:
Handled now by #293

Issue #191, opened 2014-01-29
Detach session automatically


astrodad:
Hello,

I'm looking to launch a project at startup and then detach immediately so that other people can use the session. Is there any way to do that? I tried sending in a C-b D and ^b D as the second line of the panel array but that just sent that text to the pane. Any ideas?

rightaway:
+1 for the startup option tmux_detached like in https://github.com/tmuxinator/tmuxinator/pull/284

ZeusTheTrueGod:
Guys, there is an easy trick, just add this in ```windows:``` in your .yml config file
``` -detach: tmux detach && tmux kill-window```

Soliah:
Addressed by #293

lougreenwood:
I'm not sure, but I don't know if this is really addressed.

For my use case I sometimes want to start it detached, others not, so using a `post` isn't possible because I lose optionality.

Ideally I'd prefer to just do `tmuxinator start myproject -D` and control whether to start detached from a flag...

Drew-Daniels:
@lougreenwood Agree - I don't want to have to always start a tmuxinator project detached or attached. Would be very nice if we could default to starting tmux sessions attached, but only start detached if a `--detached` or `-d` flag is provided when we  initially create the session.

PR #284, opened 2015-02-13
Add option "tmux_detached" for detach client

mvaraga:
This option specifies if tmux start with attached or detached state.
Start attached when "false" or option not defined, otherwise detached.
As requested in #191 #224

### DIFF
--- a/lib/tmuxinator/assets/sample.yml
+++ b/lib/tmuxinator/assets/sample.yml
@@ -21,6 +21,9 @@ root: ~/
 # Specifies (by name or index) which window will be selected on project startup. If not set, the first window is used.
 # startup_window: logs
 
+# for option to start detached
+# tmux_detached: true
+
 windows:
   - editor:
       layout: main-vertical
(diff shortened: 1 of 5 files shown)


maxmeyer:
@Soliah 
Does it make sense to add a "detached"-option to `start` as well?

```bash
tmuxinator start <project> --detached
```

This way one could add it to a shell initscript.
```
tmuxinator start project_1 --detached
tmuxinator start project_2 --detached
tmuxinator start project_3 --detached
tmuxinator start project_4 --detached
```

Would love to see this merged.

Soliah:
Command line flag would be great. Thor makes it relatively easy to add the option if you want to have a go @mvaraga?

mvaraga:
I do it. Should I preserve option in project file and override it with start option. Or just start option?

maxmeyer:
Personally I would expect that the --detach-option will override the configuration in the project config-file.

Soliah:
Yes sounds good that the command line option takes precedence.

paularmstrong:
This is great! Definitely a helpful addition.

maxmeyer:
@mvaraga Any progress on that? ;-) Any help needed?

Soliah:
Thinks #293 deals with this now. Can you please try and comment in that issue for followup.

PR #288, opened 2015-02-27
Removes command that causes errors

froderik:
`tmuxinator commands zsh` causes: 

    ERROR: "tmuxinator commands" was called with arguments ["zsh"]

This commit removes the extra zsh commands and solves the problem.

### DIFF
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -2,7 +2,7 @@
 
 _tmuxinator() {
   local commands projects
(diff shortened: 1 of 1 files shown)


Soliah:
This is required as the completions command outputs slightly different formatted commands for the latest completions file.

If you're using zsh-completions, those are out of date and I need to push the completions file upstream.

Soliah:
Related zsh-users/zsh-completions#299

Soliah:
Closing as this depends on above zsh-completion issue.

Issue #231, opened 2014-05-29
Add post hook


Soliah:
Add a post hook, similar to pre.

```
post:
  - tmux detach-client
```

Should deal with #224, #218, #191.

jclif:
:+1:

PR #293, opened 2015-03-21
Add post and attach config options

aMoniker:
This (partially) solves https://github.com/tmuxinator/tmuxinator/issues/231 and adds a config setting for optionally attaching/switching to sessions.

For me this is useful since I use iTerm2 which has tmux integration that works when using tmux's control mode. So now I can set my config like this:
```
attach: false
post: tmux -CC attach
```
I tried to use just `post` and call detach there, but adding anything after the attach/switch in the template means that it won't run until the user exits the session. If there's a better solution that would just use the `post` setting, I'd be happy to remove `attach`.

I've been looking for this for a while since it's annoying to have to start a tmux session, detach, then reattach using -CC manually every time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,20 @@ pre_window: rbenv shell 2.0.0-p247
 
 These command(s) will run before any subsequent commands in all panes and windows.
 
+## Custom attachment and post commands
+
+You can set tmuxiniator to skip auto-attaching to the session by using the `attach` option.
+
+```
+attach: false
+```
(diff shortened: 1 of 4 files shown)


Soliah:
Could you add something to the readme explaining this?

aMoniker:
Sure. I added some comments to the readme, and updated the sample.yml config file.

jclif:
Huzzah!

PR #295, opened 2015-03-30
Search config files in sub directories of config root

yvaucher:
Fix completion to search in sub directory of `~/.tmuxinator` (or any other root dir for config)

It is intended to provides path from root dir instead of basename to show it in completion suggestions.

Might help for #123 (#111)


yvaucher:
Thanks

PR #303, opened 2015-05-11
Doens't throw error when project file doesn't specify name

### LINKED ISSUE #302: Throws error when project file doesn't specify name
When loading a project file that doesn't specify name, following error is thrown:
```
/Users/erik/.rvm/gems/ruby-2.2.0/gems/tmuxinator-0.6.10/lib/tmuxinator/project.rb:34:in `name': undefined method `shellescape' for nil:NilClass (NoMethodError)
	from /Users/erik/.rvm/gems/ruby-2.2.0/gems/tmuxinator-0.6.10/lib/tmuxinator/project.rb:115:in `name?'
	from /Users/erik/.rvm/gems/ruby-2.2.0/gems/tmuxinator-0.6.10/lib/tmuxinator/config.rb:87:in `validate'
	from /Users/erik/.rvm/gems/ruby-2.2.0/gems/tmuxinator-0.6.10/lib/tmuxinator/cli.rb:68:in `start'
	from /Users/erik/.rvm/gems/ruby-2.2.0/gems/tmuxinator-0.6.10/bin/tmuxinator:13:in `<top (required)>'
	from /Users/erik/.rvm/gems/ruby-2.2.0/bin/tmuxinator:23:in `load'
	from /Users/erik/.rvm/gems/ruby-2.2.0/bin/tmuxinator:23:in `<main>'
	from /Users/erik/.rvm/gems/ruby-2.2.0/bin/ruby_executable_hooks:15:in `eval'
	from /Users/erik/.rvm/gems/ruby-2.2.0/bin/ruby_executable_hooks:15:in `<main>'
```

Steps to reproduce:
 * create config file with window panes, but without `name:` or `project_name:`
 * open project file
 * error will be thrown

All OSes and ruby versions affected.

underscorenygren:
the `project.rb#name` function forgot to check for nil
adds a `.blank` check similarly to how `#root` works
Adds a failing test case
Adds a test case for the same behavior for `#root` (which already works)
Fixes #302

### DIFF
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -31,7 +31,7 @@ module Tmuxinator
 
     def name
       name = yaml["project_name"] || yaml["name"]
-      name.shellescape
+      name.blank? ? nil : name.shellescape
     end
 
     def pre
--- a/spec/factories/projects.rb
+++ b/spec/factories/projects.rb
@@ -22,4 +22,12 @@ FactoryGirl.define do
 
     initialize_with { Tmuxinator::Project.new(file) }
   end
+
+  factory :noname_project, :class => Tmuxinator::Project do
+    transient do
+      file { YAML.load(File.read("#{File.expand_path("spec/fixtures/noname.yml")}")) }
+    end
+
+    initialize_with { Tmuxinator::Project.new(file) }
+  end
 end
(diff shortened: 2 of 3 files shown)


lib/tmuxinator/project.rb
```diff
@@ -31,7 +31,7 @@ def root
 
     def name
       name = yaml["project_name"] || yaml["name"]
-      name.shellescape
+      name.blank? ? nil : name.shellescape
```
maxmeyer: what about name.empty?

underscorenygren: Went with blank because that's what the #root fn did, but I don't think it
makes a difference.

PR #307, opened 2015-05-25
Feature: detached session (configuration plus command line override)

geggo98:
This is based on a similar pull request by user @mvaraga.

This adds the option for creating a session with tmuxinator without attaching to this session. The user can configure detaching in the session definition and later override the behavior on the command line with `tmuxinator start --attach=false ...` or `tmuxinator start --attach=true ...`, with `true` being the default. 

The command line always overrides the yaml file.


Soliah:
If this could be rebased ontop of #293 for the command line options that'd be great.

geggo98:
I will clean up the failed test cases later this week. I already resolved all merge conflicts.

README.md
```diff
@@ -1,3 +1,7 @@
+This is a fork of [tmuxinator/tmuxinator](https://github.com/tmuxinator/tmuxinator). In this fork I am trying to add some new features. If a feature is ready, I will send it back as a pull request.
```
dg-ratiodata: @geggo98 Was this added by mistake? I think this should not be part of a PR. :smile:

lib/tmuxinator/project.rb
```diff
@@ -4,11 +4,24 @@ class Project
...
 
-    attr_reader :yaml, :custom_name
+    attr_reader :yaml
+    attr_reader :force_attach
+    attr_reader :force_detach
+    attr_reader :custom_name
+
+    def initialize(yaml, options = {})
+      options[:force_attach] = false if options[:force_attach].nil?
+      options[:force_attach] = false if options[:force_attach].nil?
```
dg-ratiodata: Duplicate line or did you mean `#empty?`?

```
options[:force_attach] = false if options[:force_attach].nil? || options[:force_attach].false?
```

lib/tmuxinator/project.rb
```diff
@@ -56,14 +79,6 @@ def pre_window
       end
     end
 
-    def attach
-      attach = true
-      if !yaml["attach"].nil?
```
dg-ratiodata: I would use a guard clause here...

```ruby
def attach
  return yaml["attach"] unless yaml["attach"].nil?
  
  true
end
```

deivid-rodriguez: or better

```ruby
def attach
  yaml["attach"] || true
end
```

lib/tmuxinator/project.rb
geggo98: On the duplicate lines: One should be for `:force_attach` and one for `:force_detach`.

lib/tmuxinator/project.rb
dg-ratiodata: @deivid-rodriguez Can `yaml["attach"]` be `false`? Then this will not work as expected.

deivid-rodriguez: @dg-ratiodata You're totally right! I'd assume it can actually be `false` so my advice is just plain wrong. :flushed:

README.md
geggo98: My fault. This should only be part of the main branch, but never be merged to the feature branch. I wanted to have a disclaimer on the main branch of my fork to make clear it is only a fork and I did not invent tmuxinator.

Soliah:
Thanks!

dg-ratiodata:
@Soliah Would you mind to release a new gem version including this PR?

PR #311, opened 2015-06-03
Misc improvements

deivid-rodriguez:
The only improvement from the user point of view is the last commit. When no prewindow or pretab commands are defined, nothing is run instead of empty commands. This results in more available space in the screen by default.

I also added some other minor changes. Feel free to cherry pick the ones you like!

Regards.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,12 @@
 source "https://rubygems.org"
 
-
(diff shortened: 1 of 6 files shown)


Soliah:
Thanks all this looks great.

deivid-rodriguez:
You are welcome @Soliah, thanks to you for `tmuxinator`, it's great.

Issue #292, opened 2015-03-16
Ability to create unnamed windows


kennykaye:
My understanding is that currently, windows must have names. Sometimes I prefer the window name to simply be the process which is currently running.

Is there an option to do this?

youdulayo:
+1

podung:
+1.  Completely agree.

podung:
As a workaround, try this:

```
name: default
root: ~/
windows:
  - main:
    - tmux set-window-option -t1 automatic-rename on
    - clear
```

See the discussion around:  https://github.com/tmuxinator/tmuxinator/issues/174

davejamesmiller:
+1

verilucis:
:+1:

Issue #174, opened 2013-11-21
Window command in second window operates on first window


dbingham:
I have the following project:

```
# ~/.tmuxinator/dev.yml

name: dev
root: ~/

windows:
  - Main:
      layout: e1ef,425x28,0,0{111x28,0,0,0,313x28,112,0,2}
      panes:
        - tig:
          - cd /Users/dbingham/Development/git/main
          - tig --all
        - sh:
          - cd /Users/dbingham/Development/git/main
          - clear
  - dev: 
      layout: e1ef,425x28,0,0{111x28,0,0,0,313x28,112,0,2}
      panes:
        - tig:
          - tmux set-window-option automatic-rename on
          - clear
        - #bash
```

The `tmux set-window-option automatic-rename on` on the second window actually ends up setting that option on the *first* window when I run `mux start dev`.  The second window is unaffected and is still named 'dev'.

By way of workaround, is there a way to _not_ name a window when creating it?

podung:
If anyone is interested in an unnamed window option, check out https://github.com/tmuxinator/tmuxinator/issues/292.

xGabba:
@davejamesmiller keep in mind, that the windows start with 0 not with 1.

davejamesmiller:
@xGabba Not on [my machine](https://github.com/davejamesmiller/dotfiles/blob/a7526dc73f56307e60c7c6849ae3ee7d9043d093/.tmux.conf#L231) 😉

```
set -g base-index 1
set -g pane-base-index 1
```

But yes, good point. 😃

PR #323, opened 2015-07-21
292 nameless window option

podung:
Requested in issue #292 and #174.

Adds ability to create windows without a name, by using any valid yaml null key:

```yaml
name: test
root: ~/
windows:
  - ~: echo "this is a nameless window"
  - null: echo "me too"
  - Null: echo "me too"
  - NULL: echo "me too"
  - named_window: echo "this window has a name"
```

The benefit is that nameless windows use tmux's default automatic-rename behavior, meaning the displayed window name changes based on the currently running process in that window (vim, bash, tig, etc).

Let me know if I can make any changes to make this more consistent for the project.  Also, before this is merged I would like to add a note in either the README or the sample.yml file.  Please let me know where would be most appropriate.

### DIFF
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -14,7 +14,7 @@ if [ "$?" -eq 1 ]; then
 
   # Create the session and the first window. Manually switch to root
   # directory if required to support tmux < 1.9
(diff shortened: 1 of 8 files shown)


podung:
Wow, Hound is really noisy.

Some of the line length violations seem unreasonable to me, but I'm happy to address the remaining ones if you would prefer them to be fixed.

Soliah:
Thanks for this :+1: :beers:

Issue #313, opened 2015-06-12
Project names containing only numbers cause an error


rghamilton3:
Here's the config that caused an issue:
``` 
# ~/.tmuxinator/1320.yml

name: 1320 
root: ~/

startup_window: Media

windows:
  - vim: vim 
  - Omega: 
        layout: d83f,149x103,0,0[149x51,0,0,7,149x51,0,52,10] 
        panes:
           - omega
           - #empty
  - Media: ncmpcpp
  - Personal: #empty 
```

and here's the error thrown:

```
/home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:35:in `name': undefined method `shellescape' for 1320:Fixnum (NoMethodError)
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/window.rb:80:in `tmux_window_command_prefix'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/window.rb:25:in `initialize'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:24:in `new'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:24:in `block in windows'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:23:in `map'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:23:in `with_index'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:23:in `windows'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:125:in `windows?'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/config.rb:82:in `validate'
    from /home/rghamilton3/.gem/ruby/2.2.0/gems/tmuxinator-0.6.11/lib/tmuxinator/cli.rb:68:in `start'
[...]

PR #324, opened 2015-07-26
prevent project names containing only numbers from causing NoMethodError

ethagnawl:
This pull request addresses issue #313.

### DIFF
--- a/spec/fixtures/sample_number_as_name.yml
+++ b/spec/fixtures/sample_number_as_name.yml
@@ -1,0 +1,41 @@
+# ~/.tmuxinator/sample_number_as_name.yml
+# you can make as many tabs as you wish...
+
+name: 222
+root: ~/test
+socket_name: foo # Remove to use default socket
+pre: sudo /etc/rc.d/mysqld start # Runs before everything
+pre_window: rbenv shell 2.0.0-p247 # Runs in each tab and pane
+tmux_options: -f ~/.tmux.mac.conf # Pass arguments to tmux
+tmux_detached: false
+windows:
+  - editor:
+      pre:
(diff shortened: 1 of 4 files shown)


spec/fixtures/sample_number_as_name.yml
```diff
@@ -0,0 +1,41 @@
+# ~/.tmuxinator/sample_number_as_name.yml
+# you can make as many tabs as you wish...
+
+name: 222
```
podung: Would you be willing to pair this fixture down to a more bare bones tmux project file?  I think it will better isolate what your test is operating on as well as make it less likely for future changes to the parsing of the other yaml options break your test.

podung:
Thanks for this.  See my comment on the fixture file, but otherwise I'm ready to merge it.

ethagnawl:
Ok - I'll take a pass at pruning the fixture.

ethagnawl:
@podung I pruned the new fixture and made some stylistic changes per @houndci.

podung:
Looks great.  I'll merge once ci finishes up.  Thanks again.

ncoop:
I am having this problem.
tmuxinator 0.7.1
tmux 2.1

Sample yaml:
```
# ~/.tmuxinator/fixnum-test.yml
---
name: fixnum-test
root: ~/

windows:
  - 234:
```

Sample output:
```
$ mux start fixnum-test 
undefined method `shellescape' for 234:Fixnum
```

Chaining a `to_s` in `Window#initialize` seems to fix it:
```
      @name = if !window_yaml.keys.first.nil?
                window_yaml.keys.first.to_s.shellescape
              end
```

PR #326, opened 2015-07-29
add asset_path helper method to Tmuxinator::Config

ethagnawl:
`asset_path` could arguably be moved into `Tmuxinator::Util`, but seeing as it's currently only used in `Tmuxinator::Config`, it seems to make sense to keep it there as a private method.


podung:
I'll take a look at this pr this weekend.

podung:
Thanks for this.

PR #335, opened 2015-09-04
Add local project file (`pwd`/.tmuxinator.yml) support.

adamstrickland:
We have a need to be able to commit project files to git and share them
between machines (we use disposbale development boxes).  Refactored
`Tmuxinator::Config.project` to also check for `$PWD/.tmuxinator.yml`
when searching for project files.  This occurs between the recursive
search of `Tmuxinator::Config.root` and the default path being selected,
so as to allow for project files to be shared with no disruption to
existing functionality.


clekstro:
+1. The alternative workaround requires us to add our .yml files to the root via configuration managment, instead of within the project's code where it belongs.

lib/tmuxinator/config.rb
```diff
@@ -42,10 +44,21 @@ def exists?(name)
...
 
-      def project(name)
+      def project_in_root(name)
         projects = Dir.glob("#{root}/**/*.yml")
-        project_file = projects.detect { |project| File.basename(project, ".yml") == name }
-        project_file || "#{root}/#{name}.yml"
+        projects.detect { |project| File.basename(project, ".yml") == name }
+      end
+
+      def project_in_local(_name)
```
Soliah: Doesn't need `_name` right? Can't it just be removed.

Soliah:
It would be great if this could handle not having to pass in the name of the project when the file is present. So just running `mux` should start the project. This would require modification to `bin/tmuxinator` though.

Lastly when doing a `mux new` we should probably add an option (`--local`) to create the file in the current directory instead of `~/.tmuxinator`.

lib/tmuxinator/config.rb
adamstrickland: Done

adamstrickland:
@Soliah added requested `mux new --local` and no-arg `mux` to create a local project and start a local project, respectively.  Didn't see any tests to handle modifications to `bin/tmuxinator`; thougts on that?

Also, in order to decrease some commit churn I had, I have a couple of additional commits that I could queue up:
- Hound churn:
  - adds `rubocop` to `Gemfile`
  - adds a `.rubocop.yml` that, according to [Hound](https://houndci.com/configuration#ruby) (aside: I don't think that assertion is completely accurate...), is used for style-copping
- Travis churn (multi-ruby):
  - adds tasks to `Rakefile` to facilitate testing locally across all rubies defined in `.travis.yml`

I'd be happy to create PRs for either/both or just append the commits to this, at your discretion.

Soliah:
> Didn't see any tests to handle modifications to bin/tmuxinator; thougts on that?

We really need to get some integration tests setup as suggested in #336 to catch things like that.

As for Hound I'm finding it far too noisy really. What do you think about just having Rubocop run in travis instead? And having a rake task to just run rspec + integration tests if any + rubocop across Rubies would be great.

Be good to have separate PRs for whatever you want to tackle. I'll merge this PR now.

lib/tmuxinator/cli.rb
```diff
@@ -48,64 +56,93 @@ def completions(arg)
...
-      unless Tmuxinator::Config.exists?(name)
-        template = Tmuxinator::Config.default? ? Tmuxinator::Config.default : Tmuxinator::Config.sample
-        erb  = Erubis::Eruby.new(File.read(template)).result(binding)
-        File.open(config, "w") { |f| f.write(erb) }
+      project_file = if options[:local]
+                       Tmuxinator::Config::LOCAL_DEFAULT
+                     else
+                       Tmuxinator::Config.default_project(name)
+                     end
+      unless File.exists?(project_file)
```
Soliah: This should be `Tmuxinator::Config.exists?` otherwise this code block isn't entered during the tests. Not sure why travis is passing as I'm getting a failure locally here.

adamstrickland: I was unable to get this to fail when running locally.  Just to be sure, I put in a couple more tests that specifically check for the different project file paths (local or traditional).  Those are passing (locally, natch), as well.

I don't think this would mean that other `#exists?` checks would need to be changed.  My reading is that `Tmuxinator::Config.exists?` and `Tmuxinator::Config.local?` are intended to be used for checking for those specific conditions; perhaps `exists?` should be renamed `traditional?` or something like that to delineate it from `local?`.  Regardless, the `File.exists?` check here is literally a filesystem check that the selected project file path does or does not already exist, so I don't think it'd need to be changed.

adamstrickland:
WRT Hound, I agree, it's pretty noisy.  I like plain ol' rubocop as you suggested.  The rubocop configuration that Hound uses by default is a contributor to that noisiness; in particular, I think the 80-character line limit in test suites is very annoying.  I'll PR what I have as a strawman to start the discussion.

adamstrickland:
[Strawman PR for rubocop](https://github.com/tmuxinator/tmuxinator/pull/339)

mahemoff:
Doesn't #203 do the same thing?

PR #336, opened 2015-09-10
Implement passing args to YAML files

### LINKED ISSUE #242: Feature Req: Pass arguments to projects
I'm not sure if this is already supported, but I'm hoping to do things like

```
mux start ssh_run foo.bar.com
```

The first ARG is referenced within the project definition say,
```
name: sample
root: ~/

windows:
  - stats:
    - ssh $1
    - tail -f /var/log/stats.log
```

Just a thought, appreciate any feedback.
Thanks!

J3RN:
Closes #242. There's no tests currently covering how the command line actually works. I can do my best to add some, if that is preferred.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,1058 @@
+AllCops:
+  Include:
+  - "**/*.podspec"
+  - "**/*.jbuilder"
+  - "**/*.opal"
+  - "**/Vagrantfile"
+  - "**/Berksfile"
+  - "**/Cheffile"
+  - "**/Vagabondfile"
+  Exclude:
+  - "vendor/**/*"
+  - "db/schema.rb"
+  RunRailsCops: false
+  DisplayCopNames: false
+  StyleGuideCopsOnly: false
+Style/AccessModifierIndentation:
+  Description: Check indentation of private/protected visibility modifiers.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#indent-public-private-protected
+  Enabled: true
+  EnforcedStyle: indent
+  SupportedStyles:
+  - outdent
+  - indent
+Style/AlignHash:
+  Description: Align the elements of a hash literal if they span more than one line.
+  Enabled: true
+  EnforcedHashRocketStyle: key
+  EnforcedColonStyle: key
+  EnforcedLastArgumentHashStyle: always_inspect
+  SupportedLastArgumentHashStyles:
+  - always_inspect
+  - always_ignore
+  - ignore_implicit
+  - ignore_explicit
+Style/AlignParameters:
+  Description: Align the parameters of a method call if they span more than one line.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-double-indent
+  Enabled: true
+  EnforcedStyle: with_first_parameter
+  SupportedStyles:
+  - with_first_parameter
+  - with_fixed_indentation
+Style/AndOr:
+  Description: Use &&/|| instead of and/or.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-and-or-or
+  Enabled: true
+  EnforcedStyle: always
+  SupportedStyles:
+  - always
+  - conditionals
+Style/BarePercentLiterals:
+  Description: Checks if usage of %() or %Q() matches configuration.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#percent-q-shorthand
+  Enabled: true
+  EnforcedStyle: bare_percent
+  SupportedStyles:
+  - percent_q
+  - bare_percent
+Style/BracesAroundHashParameters:
+  Description: Enforce braces style around hash parameters.
+  Enabled: true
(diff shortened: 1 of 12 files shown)


magicalbanana:
Please write test. Thank you.

magicalbanana:
Do you want me to write the test for you?

magicalbanana:
It would be better also if we can pass a flag instead of just ARGS[i] so we can better specify and write our config files better.

Example:

```
mux start project vim_root=~/git/foo/mama
```

J3RN:
I'm working on a test. I also agree that the `ARGV[i]` syntax isn't ideal. I've written an alternate implementation where the configuration passed on the command line is available in an array, `args`, so you could write instead `args[0]` to get the first actual additional argument.

Having a hash with descriptive keys (e.g. `args["vim_root"]  #=> "~/git/foo/mama"`) would be quite nice, but writing out `mux start project vim_root=~/git/foo/mama` on the command line seems awfully verbose.

magicalbanana:
Great.

I get your point. But it would be nice to have that for a generic config file to be used on different projects that could be language specific or what not.

magicalbanana:
So let's say you have a vanilla config and you want to pass in the root folder at least, if you want

J3RN:
Alright, I'm starting work on the "both" feature. Will push up when it's finished.

J3RN:
`args` contains simple arguments. If you were to type "mux start foo bar", `args` would be a 1-item array, containing only "bar". `settings` contains items passed in the format `key=value`, which can be accessed with `settings["key"] #=> "value"`.

The side effect of this commit is that the session name option had to be moved elsewhere to avoid conflicting with the args. It is now in a flag called "name", which I think is a good home for it.

Example config:
```YAML
name: vim
root: ~/

windows:
  - editor: vim <%= args[0] %>
  - other: cd <%= settings["vimdir"] %>
```

This could then be called with:
`$ mux vim .vimrc vimdir=~/.vim`

lib/tmuxinator/config.rb
```diff
@@ -70,6 +82,8 @@ def validate(name, options={})
 
         config_path = Tmuxinator::Config.project(name)
 
+        settings = parse_settings(args)
```
J3RN: That's what you think. This variable can actually be accessed from within the YAML configs!

*I'm arguing with a robot CI dog*

Soliah:
Does this break #273?

J3RN:
@Soliah Yes and no. Technically, this does break reverse compatibility with that feature. However, that feature lives on in the form of a flag. If you want to specify a name, you can do so like this:
`$ mux start <project> --name=<name>`

Soliah:
Really should disable that check in hound.

bin/tmuxinator
```diff
@@ -4,16 +4,10 @@ $: << File.expand_path("../../lib/", __FILE__)
...
+name = ARGV[0]
 
-  if Tmuxinator::Cli::COMMANDS.keys.include?(name.to_sym)
-    Tmuxinator::Cli.start
-  elsif Tmuxinator::Config.exists?(name)
-    Tmuxinator::Cli.new.start(name)
-  else
-    Tmuxinator::Cli.start
-  end
+if !Tmuxinator::Cli::COMMANDS.keys.include?(name.to_sym) && Tmuxinator::Config.exists?(name)
```
Soliah: Please add a guard against a `nil` name here. Running `mux` should just show the help.

Soliah: Actually the check for an `ARGV.length == 1` should probably be there instead.

Soliah:
Could you also add stuff to the readme describing this feature in the ERB section. Example usage would also be good.

maxmeyer:
Suggestion: Why not adding some `cucumber` tests + `aruba` and use that combination to document the behaviour of tmuxinator? The rspec and the cucumber guys use that kind of pattern to document the public api for their users. I also started to move to that kind documentation for the `aruba` project itself. Maybe that's also worth for you to try. You could even use http://relishapp.com to publish this - and don't worry that one is free for public projects.

bin/tmuxinator
J3RN: I added `name &&` to make sure `nil` wouldn't get through. Good catch!

J3RN:
@Soliah Just updated the README. Let me know if it needs more work.

J3RN:
@maxmeyer That's a pretty cool idea. I'm feeling like it's a little out-of-scope here, though. I'd say you should open an issue for that, and we can do that in another PR. That is, unless the maintainers think that we should include it here.

Soliah:
I just merged #335 which was a rather large change. Can you please rebase on master?

J3RN:
On it.

magicalbanana:
Yeah, you need to rebase.

J3RN:
It looks like force pushing to my branch closed the PR? Can one of you fine gentlemen re-open it? @magicalbanana @Soliah

J3RN:
Nevermind. It looks like GitHub is all sorts of confused. I opened a new PR at #343.

PR #339, opened 2015-09-23
Adds rubocop

adamstrickland:
- Adds `rubocop` gem
- Adds `.rubocop.yml` configuration
- Adds a couple of rake tasks to run them similar to how Hound does (i.e. only cop'ing the files that were touched)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -506,6 +506,8 @@ Metrics/LineLength:
   URISchemes:
   - http
   - https
+  Exclude:
+    - lib/tmuxinator/window.rb
 Metrics/MethodLength:
   Description: Avoid methods longer than 10 lines of code.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#short-methods
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,1056 @@
+AllCops:
+  Include:
+  - "**/*.podspec"
+  - "**/*.jbuilder"
+  - "**/*.opal"
+  - "**/Vagrantfile"
+  - "**/Berksfile"
+  - "**/Cheffile"
+  - "**/Vagabondfile"
+  Exclude:
+  - "vendor/**/*"
+  - "db/schema.rb"
+  RunRailsCops: false
+  DisplayCopNames: false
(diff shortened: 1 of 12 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,1056 @@
+AllCops:
```
adamstrickland: According to the [Hound website](https://houndci.com/configuration#ruby), this is the rubocop configuration being used.  However, I don't think that's entirely true...  as several cops defined here did not report by Hound.  I have commented out a few cops (commenting them, rather than simply removing them, so that we can see what they are) that we just not working.

Gemfile
```diff
@@ -5,8 +5,9 @@ gemspec
...
-gem "simplecov", "~> 0.9"
-gem "coveralls", "~> 0.7"
-gem "awesome_print", "~> 1.2"
-gem "pry", "~> 0.10"
+gem 'simplecov', '~> 0.9'
+gem 'coveralls', '~> 0.7'
+gem 'awesome_print', '~> 1.2'
+gem 'pry', '~> 0.10'
 gem "factory_girl", "~> 4.5"
+gem 'rubocop', "~> 0.34", require: false
```
Soliah: Looks good. Please use double quotes here.

Soliah:
What do you think about running Rubocop in Travis instead of using hound?

adamstrickland:
Cleaned up all rubocop warnings so, yes, rubocop could now be added to travis.

lib/tmuxinator/wemux_support.rb
```diff
@@ -5,17 +5,17 @@ def wemux?
     end
 
     def load_wemux_overrides
-      self.instance_eval do
-        def render
+      class_eval do
```
adamstrickland: Had to change this around due to the `Lint/NestedMethodDefinition` cop which didn't like the `def`s inside a method definition.  Added a test to make sure that the functionality stays consistent, but if could also just revert this and deactivate the cop itself.

Soliah:
Great. You might as well add rubocop to `.travis.yml` and I'll go ahead and remove hound.

adamstrickland:
Done

Issue #253, opened 2014-09-27
tmux global settings support


azat:
Right now there is no way to install tmux settings per session (tmuxinator session).
It is useful for `history-limit`, because it will be applied only for newly created windows, not for existing one (hence `pre_window` won't work)

Soliah:
Try setting `tmux_options` to `-s set history-limit 1`.

Soliah:
Actually that won't work.

Soliah:
I think you should be able to put this in the newly supported `post`. #293.

azat:
On Sat, Jun 06, 2015 at 05:23:23AM -0700, Christopher Chow wrote:
> I think you should be able to put this in the newly supported `post`. #293.

If I understand correctly, then it will also won't work.

Since you can't change history-limit when window/pane already created,
and patch from #293 just allow you to execute some command *after*
windows/panes already created.

Am I missing something?

azat:
Actually I can't even use `pre` since `tmuxinator` creates server before:
```
tmux -f /tmp/tmux.mail.conf start-server\; has-session -t mail 2>/dev/null

if [ "$?" -eq 1 ]; then
  ...
  # Run pre command.
  echo set history-limit 1000 >> /tmp/tmux.mail.conf
```

ethagnawl:
Does anyone else think this is worth pursuing? Aside from `history-limit`, are there any other scenarios where this option would be useful?

It'd be trivial to add a `pre_tmux` option. Though, the naming might be a bit confusing as there's already a `pre` option.

azat:
@ethagnawl can this [hunk](https://github.com/azat/tmuxinator/commit/e2ce490197d2b814fb14dc5505bd347cd8f8a5cb#diff-fd1cfbe648a569f6ca3d8703d96ceb22R101) at least be applied?

Otherwise I can't use `bash` specific stuff in `pre`

ethagnawl:
@azat I'd prefer introducing the explicit `pre_tmux` option. Does anyone else have thoughts either way? (cc @Soliah, @J3RN, @adamstrickland)

azat:
> @azat I'd prefer introducing the explicit pre_tmux option. Does anyone
else have thoughts either way? (cc @Soliah, @J3RN, @adamstrickland)

But I still need to change default interpreter.

Thanks, Azat.

hrchu:
I need this. Any workaround?

PR #340, opened 2015-09-30
Use SHELL from env

azat:
This is a temporary fix of #253 for me, since this patch will allow to use bash
process substitution to extend tmuxinator:

```
tmux_options: -f <(echo set -g history-limit 1000)
```

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -123,11 +123,11 @@ module Tmuxinator
           STDIN.getc
         end
 
-        Kernel.exec(project.render)
+        Kernel.exec(ENV["SHELL"] || "/bin/sh", "-c", project.render)
       end
 
       def kill_project(project)
-        Kernel.exec(project.tmux_kill_session_command)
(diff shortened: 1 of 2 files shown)


J3RN:
@azat Hey there! Thanks for your pull-request! :sparkles: I'm sorry that it's taken over a year for it to be reviewed.

I absolutely agree that Tmuxinator should use the user's configured shell. To take this PR forward, it looks like the first thing is that you'll need to rebase in recent changes on master. Would you mind doing that?

azat:
Hi @J3RN, sure, push forced to this branch
Thanks!

adamstrickland:
Unfortunately, this has sat neglected for a long time, and *I think* is no longer relevant...  the templates dictate using the user's shell (e.g. https://github.com/tmuxinator/tmuxinator/blob/master/lib/tmuxinator/assets/template.erb#L1), so we're ok there...  the use of the user's shell in `Tmuxinator::Project#extract_tmux_config` fails w/ the suggested edit.  I'm going to close this PR for now, but we can re-open if there's still an issue.

PR #356, opened 2015-11-16
Add delete multiple projects feature

### LINKED ISSUE #337: Delete multiple projects at once
This is a feature request. I'd love to be able to delete multiple projects with a command like:
```bash
mux delete project1 project2 ...
```

J3RN:
This feature allows you to delete multiple projects by typing:
```bash
mux delete <PROJECT1> <PROJECT2>
```

It's not super urgent, but it's a nice convenience. Added tests. Closes #337.

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -180,20 +180,22 @@ module Tmuxinator
       Kernel.system("$EDITOR #{new_config_path}")
     end
 
-    desc "delete [PROJECT]", COMMANDS[:delete]
+    desc "delete [PROJECT1] [PROJECT2] ...", COMMANDS[:delete]
     map "d" => :delete
     map "rm" => :delete
 
-    def delete(project)
-      if Tmuxinator::Config.exists?(project)
-        config = "#{Tmuxinator::Config.root}/#{project}.yml"
+    def delete(*projects)
+      projects.each do |project|
+        if Tmuxinator::Config.exists?(project)
+          config = "#{Tmuxinator::Config.root}/#{project}.yml"
 
-        if yes?("Are you sure you want to delete #{project}?(y/n)", :red)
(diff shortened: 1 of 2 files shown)


ethagnawl:
This isn't necessarily a problem with your PR, as the issue was pre-existing, but it looks like `config = "#{Tmuxinator::Config.root}/#{project}.yml"` ignores local config files.

ethagnawl:
@J3RN As it's not directly related to your changes, I've created an [issue](https://github.com/tmuxinator/tmuxinator/issues/390) for the issue noted above and will address it in a separate PR.

One nitpick I noticed while re-reviewing these changes just now: should `say` be used instead of `puts` in [`Cli::delete`](https://github.com/tmuxinator/tmuxinator/pull/356/files#diff-fd1cfbe648a569f6ca3d8703d96ceb22R197)? IIRC, it's just an alias for `puts` and they're both used throughout (I wouldn't be opposed to opening an issue for this), but this one stuck out because `say` is used in the other branch of the containing conditional.

Otherwise, this looks good to me. :+1:

J3RN:
@ethagnawl Thanks for doing the separate issue for that one! I had been meaning to get around to it, but never did.

I think I fixed what you were talking about. Good call on the `say` thing -- I had never actually noticed that `say` was a real method.

ethagnawl:
> I think I fixed what you were talking about. Good call on the say thing -- I had never actually noticed that say was a real method.

Looks good! So, [`say`](http://www.rubydoc.info/github/wycats/thor/Thor%2FShell%2FBasic%3Asay) is actually provided by Thor and it does have some additional capabilities and behaves slightly differently than `puts`. I think it's probably worth creating an issue about standardizing its use.

ethagnawl:
@adamstrickland Shall we merge this one?

Issue #201, opened 2014-02-10
Set root from environment variable


mahemoff:
Is there any way to set the `root` field from an environment variable? It would be a useful feature request if not, so project collaborators can share config while running from different paths.

PR #357, opened 2015-11-17
347 - prevent tmux from opening sessions using prefixes

### LINKED ISSUE #347: Naming with namespace cause bug when start  a project
I have a project called `apriori` and the other called `apriori-test`. The issue is that after I started `apriori-test` and want to start the `apriori` project, it will open the `apriori-test` session instead. But if I started `apriori` first and want to start `apriori-test`, it works without any issue. And the `new` command also works fine.
How can I solve this?

ethagnawl:
Fixes #347, wherein calling `mux project` would attach to an existing session called `project-test`.

I've been scratching my head about how to write a failing test that this PR will make pass, but haven't come up with anything yet. Does anyone have any ideas?

Also, HT to @nicm for suggesting the use of `has-session = ...` in [Tmux issue #201](https://github.com/tmux/tmux/issues/201).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,4 +10,4 @@ gem "coveralls", "~> 0.7"
 gem "awesome_print", "~> 1.2"
 gem "pry", "~> 0.10"
 gem "factory_girl", "~> 4.5"
-gem "rubocop", "~> 0.34", require: false
+gem "rubocop", "0.35.1", require: false
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -4,7 +4,7 @@
 unset RBENV_VERSION
 unset RBENV_DIR
 
-<%= tmux %> start-server\; has-session -t <%= name %> 2>/dev/null
+<%= tmux %> start-server\; has-session = <%= name %> 2>/dev/null
 
 if [ "$?" -eq 1 ]; then
   cd <%= root || "." %>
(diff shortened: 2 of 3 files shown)


podung:
We don't really have a great strategy for writing tests around the template.  I think two valid approaches are:
1) shove the contents of that line into a method on Project, where it can be spec'd there.  Then just invoke it in the template for now without spec'ing it there.
2) Do a spec at the project level calling `#render` and assert that it contains the one line you expect.

Any other ideas?  I think one of these would suffice for now.

adamstrickland:
Of the two suggestions, I think #2 is probably the better test scenario, as it most closely captures the intent: "insure that the resulting template uses a particular syntax".

ethagnawl:
@podung @adamstrickland Thanks for your feedback. I'll try to put something together this week.

ethagnawl:
It looks like there's [a Rubocop configuration issue](https://travis-ci.org/tmuxinator/tmuxinator/jobs/103280073#L1088) which is causing the build to fail.

```
Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.
```

Do we want to try to address that in this PR? `Style/TrailingComma` is currently disabled, so we _should_ be able to drop in either/both of those suggestions.

adamstrickland:
I think we have 2 options here:

- pin `rubocop` to `0.35.1` (the commit that broke this is in `0.36.0`), or
- upgrade the cop to the new ones and pin to `0.36.0` (to prevent similar future issues).

As far as I'm concerned, either is acceptable.  If I had my druthers, I'd probably prefer the former.

ethagnawl:
@adamstrickland I'm fine with pinning Rubocop to 0.35.1 to get this PR merged. I'd be happy to open a new issue about upgrading, if you (or anyone else) thinks it'd make sense.

adamstrickland:
Yeah, let's do that

PR #358, opened 2015-11-18
Editing will not recreate projects

### LINKED ISSUE #354: Editing project recreates it
When I run `tmuxinator edit projectname` it recreates the project yml file with the default one.

I believe the bug is on this line:

https://github.com/tmuxinator/tmuxinator/blob/master/lib/tmuxinator/cli.rb#L75

^the function expects project name, not the path.

### LINKED ISSUE #355: mux new <name> erase existing project
It's unexpected to erase existing projects. It should give an error or fallback to edit mode.
v0.7.0

J3RN:
I have re-written a few tests to better test their functionality. Additionally, I have build and installed this branch of the gem to ensure it actually works now.

@adamstrickland Thoughts?

Closes #355
Closes #354

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -72,7 +72,7 @@ module Tmuxinator
                      else
                        Tmuxinator::Config.default_project(name)
                      end
-      unless Tmuxinator::Config.exists?(project_file)
+      unless File.exists?(project_file)
         template = Tmuxinator::Config.default? ? :default : :sample
         content = File.read(Tmuxinator::Config.send(template.to_sym))
         erb = Erubis::Eruby.new(content).result(binding)
--- a/spec/lib/tmuxinator/cli_spec.rb
+++ b/spec/lib/tmuxinator/cli_spec.rb
@@ -164,7 +164,7 @@ describe Tmuxinator::Cli do
 
       context "existing project doesn't exist" do
         before do
-          expect(Tmuxinator::Config).to receive_messages(exists?: false)
+          expect(File).to receive_messages(exists?: false)
         end
 
         it "creates a new tmuxinator project file" do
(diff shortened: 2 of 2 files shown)


adamstrickland:
Does this fix the bug described?  FWIW, I had a few minutes the other night and was not able to conclusively reproduce the reported failure behavior within a test; I don't see in this changeset anything along those lines?

adamstrickland:
As for replacing `Tmuxinator::Config.exists?` with `File.exists?`...  In general I prefer to use system libraries such as `File` rather than custom abstractions.  However, the use of `Tmuxinator::Config.exists?` predates my association with the project, so there may well be reasoning for it of which I am not aware.  @aziz @Soliah may have additional context.

J3RN:
The big issue here is that `Tmuxinator::Config#exists?` expects a *project's name* to be passed, but what is being passed here is a *full file path to the project file*. `Tmuxinator::Config#exists?` treats this full file path as the project's name, and cannot find such a project. Thus it "creates it", which incidentally overwrites the already-existing one. This is easily mitigated by replacing `Tmuxinator::Config#exists?` in this instance with `File#exists?`, which does expect a full file path.

J3RN:
WRT the failing tests, it is only failing for TMUX master, which unfortunately [the `master` branch is as well](https://travis-ci.org/tmuxinator/tmuxinator?branch=master).

adamstrickland:
@J3RN I wanted some tests for this specific failing case so that it doesn't slip in again (one of the shortcomings of using mocks, eh?), so I created some, merged your branch on top and now we've got green again.  To wrap it all together, I created a separate [PR](https://github.com/tmuxinator/tmuxinator/pull/359) (side note: anyone know the proper git-iquette for this scenario?).  What do you think?

Issue #372, opened 2016-01-26
Feature Request: Custom [default].yml


JohnRodney:
Hi, I have been working on my local installed version of tmuxinator and setting up a way for me to pass a template for it to use rather than default.yml if it exists.  

Use Case: I have multiple projects that match a similar structure and will need to execute commands on launch to run servers ect. Using default.yml as that project setup is nice, but I also have other projects that match setup that aren't like the default.yml.  What I would like is a syntax to pass the template to use for its default.

This is a feature I will be adding to my downloaded version and was wondering if it was something the community was interested in and if so what would the preferred syntax for passing in the default look like?

I don't think the change would be too major as the File system already has a method exists which could be passed the filename (which it uses to check if it should open the file or create a new yml) and it already has a CMD line parser using THOR.  So it would just be passing the choosen syntax into the exists function then using that as [default] rather than default.yml.

JohnRodney:
I have it working on my local with this syntax
`tmuxinator open [projectName] --default [defaultName]`

It then checks if the file exists if it does not then it falls back to default if this doesn't exist then it uses the sample provided.

Open to suggestions.

PR #373, opened 2016-01-26
Adds a default argument for a --default [filename].yml to be used as the

JohnRodney:
template.

The use case for this is when having muliple simular projects in your
daily tasks that can't inherit from only one template type.

The approach used was to dry up code by removing default? which is
essentially exists?(filename) but without accepting the argument.
Therefore this method could be removed. default(filename) was adjusted
to accept a filename parameter which still defaults to "default" if one
isn't passed.  Finally if filename and default do not exist then the
sample method should be returned.  This also removed the need for
calling to_sym as the default already returns the appropriate file
path.

### DIFF
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -14,12 +14,11 @@ module Tmuxinator
         asset_path "sample.yml"
       end
 
-      def default
-        "#{ENV['HOME']}/.tmuxinator/default.yml"
-      end
-
-      def default?
-        exists?("default")
+      def default(filename)
+        if exists? (filename)
+          return "#{ENV['HOME']}/.tmuxinator/#{filename}.yml"
+        end
+        sample
       end
 
       def installed?
--- a/spec/lib/tmuxinator/config_spec.rb
+++ b/spec/lib/tmuxinator/config_spec.rb
@@ -15,6 +15,7 @@ describe Tmuxinator::Config do
 
   describe "#default" do
     it "gets the path of the default config" do
+      allow(Tmuxinator::Config).to receive(:exists?).and_return(true)
       expect(Tmuxinator::Config.default("default")).to include("default.yml")
     end
     it "maps the path of sample if default fails" do
--- a/spec/lib/tmuxinator/config_spec.rb
+++ b/spec/lib/tmuxinator/config_spec.rb
@@ -21,7 +21,6 @@ describe Tmuxinator::Config do
     it "maps the path of sample if default fails" do
       expect(Tmuxinator::Config.default("notafile")).to include("sample.yml")
     end
-
   end
 
   describe "#default_path_option" do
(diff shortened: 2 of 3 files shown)


JohnRodney:
This is in reference to open issue feature request #372

Chun-Yang:
+1

J3RN:
Sorry for the delay on this! I will test it out and give you feedback.

JohnRodney:
@J3RN no worries as I had mentioned I hacked my local version to have this behavior so its not a big deal either way.  Just wanted to share the solution in case anyone else had my situation where you end up bootstrapping multiple project types like Meteor Projects, Ruby Projects, Node Projects, ect.  

Its been quite a while since I wrote this code so may take a refresher if you request me to alter it in any way.

JohnRodney:
@J3RN looks like there are some merge conflicts now.  Did yo get a chance to look at this / have any feedback or requested updates to the code for me?

J3RN:
@JohnRodney I apologize for the delay. I have pulled this down, verified that it works, and reviewed the code. My only feedback is nitpicks, and I will leave comments on the appropriate lines.

lib/tmuxinator/config.rb
```diff
@@ -14,12 +14,11 @@ def sample
...
       end
 
-      def default
-        "#{ENV['HOME']}/.tmuxinator/default.yml"
-      end
-
-      def default?
-        exists?("default")
+      def default(filename)
+        if exists? (filename)
```
J3RN: For consistency with the rest of the project, there should be no space after the `?`.

spec/lib/tmuxinator/config_spec.rb
```diff
@@ -15,7 +15,11 @@
 
   describe "#default" do
     it "gets the path of the default config" do
-      expect(Tmuxinator::Config.default).to include("default.yml")
+      allow(Tmuxinator::Config).to receive(:exists?).and_return(true)
+      expect(Tmuxinator::Config.default("default")).to include("default.yml")
+    end
+    it "maps the path of sample if default fails" do
```
J3RN: Again, for consistency, there should be an empty line above this one (in between `end` and `it`.

J3RN:
:+1: with rebase and comments.

lib/tmuxinator/config.rb
ethagnawl: Perhaps this should also become an if/else?

J3RN: The one-liner syntax that RuboCop adores would have this method look like the following, since it generally sides against having methods be only an if/else statement:

```ruby
      def default(filename) 
        return "#{ENV['HOME']}/.tmuxinator/#{filename}.yml" if exists?(filename)
        sample
      end
```

At least by default, that is. We may have customized RuboCop.

JohnRodney:
thanks @J3RN I will look into making these adjustments in the next day or so.

J3RN:
@JohnRodney Any updates on this?

J3RN:
@JohnRodney Are you still interested in getting this merged in?

PR #382, opened 2016-03-04
Downgraded simplecov for coveralls compatibility

J3RN:
On my machine, running `bundle install` resulted in Bundler complaining that our `Gemfile` asked for version 0.11.0 or simplecov, while our `Gemfile.lock` had it locked in at version 0.10.0. After trying to upgrade simplecov, it turns out that the latest version of coveralls requires the outdated version of simplecov. So I modified the Gemfile to say to use version 0.10.0, and all should be well and good.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gemspec
 
 gem "rake", "~> 10.4"
 gem "rspec", "~> 3.3"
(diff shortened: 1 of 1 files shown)


J3RN:
Whoops, I just realized we don't check in `Gemfile.lock`. My mistake.

PR #383, opened 2016-03-06
Use `tmux -V` command for checking tmux existence

aycabta:
Windows has `type` command, but it does not work the same as Unix like OS's one, it's similar to `cat` command. So, checking tmux existence failed on Windows (MSYS2) with `type` command.

`tmux -V` returns 0 when `tmux` command exists, otherwise non-0 be returned. This is equivalent of `type tmux` on Unix like OS, for checking `tmux` command existence.

### DIFF
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -23,7 +23,7 @@ module Tmuxinator
       end
 
       def installed?
-        Kernel.system("type tmux > /dev/null")
+        Kernel.system("tmux -V > /dev/null")
       end
 
       def version


J3RN:
@aycabta Thanks for your PR! :sparkles:

I guess my biggest concern about this would be that since `/dev/null` does not exist on Windows, an ugly error message will likely be printed to the console (granted, I haven't used Windows in several years).

While it's a little more complex, I would think that it would make sense to run a different command if the user is running Windows vs a *nix-like system. A brief Google turns up the `Get-Command` command, which might be best for this purpose. Would you like to try implementing that?

adamstrickland:
I think @J3RN's suggestion to use `Get-Command` is probably the right way to go, but like him I haven't used Windows in years.  Will need someone to take this on.

Closing due to inactivity.

Issue #345, opened 2015-10-09
_tmuxinator:2: command not found: tmuxinator


WhiskeyTuesday:
When I open a new terminal I get
`_tmuxinator:2: command not found: tmuxinator`
`_tmuxinator:3: command not found: tmuxinator`
printed at the top.

I'm running:
Terminator
Debian Jesse (Actually Crunchbang ++)
RVM
Standard ruby 2.2.1
ZSH

This occurs whether I have a tmuxinator project in the list or not. if I comment out the line in my zshrc that sources the completion script it goes away.

blainekasten:
I'm getting this also. It happened to me after updating to OS X El Capitan

blainekasten:
Reinstalling worked for me. Did you try that @WhiskeyTuesday ? `gem install tmuxinator`

iasj:
Not working for me too. Even after reinstall.

iasj:
Forget it. I just made my own initializer [IniTmux](https://gitlab.com/iasj/IniTmux).

eliasbagley:
This is happening to me too. Tmuxinator is installed, and reinstalling isn't working for me.

magicalbanana:
Through El Kapitan? Can you tell me what ruby versioning system you guys are using? Because there's an issue with El Kapitan and rbenv.

hetre85:
Happening for me too:

Yosemite
RVM
zsh
iTerm2

Tried `gem uninstall` and a subsequent reinstall of the gem.

I am still able to use `tmuxinator [project]`, I am just receiving the message `_tmuxinator:2: command not found: tmuxinator` when I start iterm2.

Soliah:
This looks like an issue with completions. Are people using zsh-completions?

WhiskeyTuesday:
I'm using Prezto, which I believe includes zsh-completions

eliasbagley:
Unless it's zsh-completions is being included in some other way, I don't think I'm using it. I don't have it listed in my .oh-my-zsh plugins folder.

hetre85:
I'm using oh-my-zsh but I am not using zsh-completions.

kfrz:
I'm also getting  `_tmuxinator:2: command not found tmuxinator` and `_tmuxinator:3: command not found: tmuxinator`. Running: 

El Capitan
RVM (ruby 2.0.0)
oh-my-zsh 
iterm2

zkilgore:
[...]

PR #386, opened 2016-03-24
Completions: test if tmuxinator is actually on $PATH

wildefyr:
I'm sure others carry their shell environment around as well on to machines where they don't want to install tmuxinator. As such trying to source the completion script results in the annoying message of 'tmuxinator: not found'. If I want tmuxinator, I'll install it if I need it.

### DIFF
--- a/completion/tmuxinator.bash
+++ b/completion/tmuxinator.bash
@@ -8,7 +8,7 @@ _tmuxinator() {
     if [ "$COMP_CWORD" -eq 1 ]; then
         local commands="$(compgen -W "$(tmuxinator commands)" -- "$word")"
         local projects="$(compgen -W "$(tmuxinator completions start)" -- "$word")"
- 
+
         COMPREPLY=( $commands $projects )
     elif [ "$COMP_CWORD" -eq 2 ]; then
         local words
@@ -21,4 +21,6 @@ _tmuxinator() {
     fi
 }
 
-complete -F _tmuxinator tmuxinator mux
+type tmuxinator 2>&1 > /dev/null && {
+    complete -F _tmuxinator tmuxinator mux
+}
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -19,7 +19,9 @@ _tmuxinator() {
   return
 }
 
-_tmuxinator
+type tmuxinator 2>&1 > /dev/null && {
+  _tmuxinator
+}
 
 # Local Variables:
 # mode: Shell-Script


adamstrickland:
:+1: 

any other opinions?

J3RN:
I saw this and am in favor :+1:

cinaeco:
Er... this change has been a bit annoying. When using `bash`, haven't we just traded one error message for another? I get `-bash: type: tmuxinator: not found`. It used to be that `complete -F _tmuxinator` would not throw an error, but `type` now does.

cinaeco:
also, if we're talking about zsh, perhaps we should get people to add the tmuxinator completions via zsh's `fpath` (see my comment in #345), rather than sourcing it? That would mean none of these "does tmuxinator exist" checks would be necessary.

J3RN:
This is probably what I deserve for not testing it myself. The output from `type` *should* be redirected to `/dev/null`, and thus not displayed, but I guess that's not happening?

cinaeco:
That is probably so, but I think even more than that, it is unnecessary to test for tmuxinator's existence before running bash's `complete -F`. Bash's completion will run `_tmuxinator` only when tmuxinator attempts tab completion... but tmuxinator isn't installed, it will just never call that, so there's no error anyway.

At least for bash, I think tmuxinator.bash should be reverted.

cinaeco:
Similar zsh: if you use the `fpath` completion loading method, you don't need the existence check - zsh will load tmuxinator.zsh/_tmuxinator as and when completion is invoked.

J3RN:
@cinaeco I opened a separate issue for this, see #395

wildefyr:
Odd, I choose to use `type` because it should be implemented according to POSIX. Clearly whatever bash version you are running doesn't do this.

cinaeco:
It does the same on two bashes: Apple's built-in 3.2+ and Centos's built-in 4.2+... I don't think this is an issue of posix compliance. `type` is fine. I personally like `command -v` better, which is also posix compliance. The 2 issues with this pull request I see are:

1. It seems stderr is wrongly redirected - it was intended to be sent to /dev/null, but isn't. That's easy to fix, but...
2. A `tmuxinator` existence check is just plain unnecessary, because `complete` works fine without it. At least, I don't see any cons. Not having the existence check makes tmuxinator.bash More Flexible because it means `tmuxinator` actually doesn't have to be detectable at time of sourcing in order for completion to work.

wildefyr:
Yes I completely agree with number 2. I wasn't aware that the bash completion completely worked fine when I implemented this: I admitted only checked zsh completion with **did** throw an error when trying to source the completion. I've been meaning to look into zsh completion for sometime now; I think switching to using fpath might be the best solution.

cinaeco:
Luckily `tmuxinator.zsh` can still have the existence check, because it will work whether it is placed in fpath` or sourced directly (though sourcing directly is still less flexible). I think I'll just create a pull request to at least revert `tmuxinator.bash`.

PR #396, opened 2016-04-01
Fix bash completion error

### LINKED ISSUE #395: Bash error when loading tmuxinator completion
It appears that the following line doesn't work as expected in Bash:
```bash
type tmuxinator 2>&1 > /dev/null
```
Instead of redirecting the error to `/dev/null` when tmuxinator is not installed, it is printed to the console, which is annoying.

The following line should work instead:
```bash
type tmuxinator &> /dev/null
```

J3RN:
By using "all output redirection", bash should no longer print the error to
console when tmuxinator is not installed. The zsh checking has been updated to
use the same method for parity.

Closes #395

### DIFF
--- a/completion/tmuxinator.bash
+++ b/completion/tmuxinator.bash
@@ -21,6 +21,6 @@ _tmuxinator() {
     fi
 }
 
-type tmuxinator 2>&1 > /dev/null && {
+type tmuxinator &> /dev/null && {
     complete -F _tmuxinator tmuxinator mux
 }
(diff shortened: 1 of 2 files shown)


cinaeco:
Following the discussion in #386, could we actually just remove the bash check and keep the zsh check?

wildefyr:
+1 to keep the zsh check for now; at least until we can make it use fpath or something similar.

J3RN:
I am closing this in favor of #400

PR #400, opened 2016-04-02
Improve bash/zsh completion-setup flexibility

cinaeco:
By removing the `tmuxinator`-existence checks and by employing the
correct completion definition functions, `tmuxinator.bash` and
`tmuxinator.zsh` can be sourced at any time, without running into errors
caused by not-yet-initialised `rvm` installations.

For `zsh`:

Previously, the first `#compdef` line was inserted that allowed
`tmuxinator.zsh` to be used within `zsh`'s completion autoloading,
however, since it seems preferable to install `tmuxinator` completion by
directly sourcing files, this autoload-related definition is no longer
necessary, and an explicit `compdef` takes its place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Bugfixes
 
 - Bugfix for issue with using numbers as window names
-- Bugfix for trying to load shell completions before checking to see if tmuxinator is installed.
+- Bugfix for zsh-completion loading throwing an error if tmuxinator is not yet available.
 - Bugfix for using `mux delete` to delete local projects
 
 ## 0.7.2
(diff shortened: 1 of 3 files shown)


cinaeco:
@J3RN this is my take on #396

cinaeco:
Of course, this doesn't solve a user typing `mux<Tab>` on systems without  `tmuxinator`, but I think in that case having the shells throw errors is appropriate.

cinaeco:
Hmmm... actually "sourced anytime" after completion initialisation... but I think the idea is there.

cinaeco:
Hmmm... that should actually be "sourced at any time" *after completion initialisation*... but I think the idea is there.

cinaeco:
Haha... er... https://github.com/zsh-users/zsh-completions/blob/master/src/_tmuxinator
Seems some people do think `$fpath` is the way to go.

J3RN:
Hey @cinaeco! Thanks for this pull request! I'm no zsh hacker, so this decision might be better left to someone better informed. However, I like the clean, simplistic approach taken here.

adamstrickland:
So are we good to :ship: this?  LGTM

J3RN:
I tested it out, and this seems to work well.

@cinaeco Would you mind throwing a note in the CHANGELOG.md under "Unreleased"? I think this falls under "Bugfixes". Then we'll merge!

cinaeco:
Updated the changelog line describing the previous insertion of the existence checks.

J3RN:
@cinaeco Looks good to me! Merging.

PR #403, opened 2016-04-04
Updated "has-session" check

Richard-Walton:
has-sesson check which does not match prefixes and works with tmux 1.8 >

### DIFF
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -4,7 +4,7 @@
 unset RBENV_VERSION
 unset RBENV_DIR
 
-<%= tmux %> start-server\; has-session = <%= name %> 2>/dev/null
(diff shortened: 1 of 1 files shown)


J3RN:
Hey again @Richard-Walton! Thanks for moving this to a PR. It looks like the failing test simply checks the contents of the template file and expects to find "has-session =", which it naturally does not. I'm not sure how we could write a better, less implementation-dependent test for this, but would you mind updating that test to pass?

Soliah:
I'd be hesitant to use `grep` as historically using platform dependent things causes bugs for people. We're probably better off potentially doing this check in Ruby. #171 is an example of where this was an issue and was fixed by doing checks in Ruby.

Richard-Walton:
@Soliah - Agreed.  Hold on.

Richard-Walton:
New pull request @ https://github.com/tmuxinator/tmuxinator/pull/404

Issue #402, opened 2016-04-04
No longer attaches to existing session


Richard-Walton:
This commit (https://github.com/tmuxinator/tmuxinator/commit/452d3bd6135631b58af9f75599b8899ebaec63d2) seems to generate an invalid command on my machine.  The commit is from Nov 2015(!) though, which seems super strange to start causing issues.  Has this code just made it into a build?

Anyway, the command (After removing the >/dev/null) fails with:

`usage: has-session [-t target-session]`

Causing the exit code to be 1.  This then means tmuxinator will create the windows/panes again every time :(

Richard-Walton:
Ok - Further investigation reveals that the `has-session = ...` syntax is not available in tmux1.8 (The version I have installed from ubuntu/trusty repos)

I tried upgrading to tmux v2.0 - the problem still persists.

Richard-Walton:
Downgrading to 0.7.1 fixes the issue

J3RN:
@Richard-Walton Hey there! I'm sorry about this issue you've been having. The change was part of a feature/fix where tmuxinator would formerly attach by prefix -- so if you has a session named "foobar" and ran `mux foo`, you'd be attached to the "foobar" session.

I'm not sure why upgrading to 2.0 didn't help, but I'm currently running 2.1, so it may have been added in that release. In any event, I can imagine that Ubuntu users are probably a large portion of our user base, so getting this fixed is a priority.

In the tmux man page, I found:
> If the session name is prefixed with an '=', only an exact match is accepted (so '=mysess' will only match exactly 'mysess', not 'mysession').

Could you try running something like:
```bash
tmux has-session -t =foo
```
and post the output?

Richard-Walton:
@J3RN - No problem.  Thanks for maintaining such a great tool.

With tmux 1.8 I have a session named "dev-env" running:

```
tmux ls
dev-env: 18 windows (created Mon Apr  4 12:51:57 2016) [166x51]

tmux has-session -t dev-env && echo $?
0
```

Unfortunately adding = as you suggested does not work:

```
tmux has-session -t =dev-env
session not found: =dev-env
```

Richard-Walton:
I wonder if perhaps, to maintain compatibility with older versions of tmux, we could do something like:

`tmux -s | grep "^<% name %>:"`

?

ethagnawl:
@Richard-Walton My apologies - I didn't realize that syntax wasn't backwards-compatible. 

[...]

PR #404, opened 2016-04-05
Updated "has session" check

Richard-Walton:
Fix for: https://github.com/tmuxinator/tmuxinator/issues/402

Also, improved template rendering to eliminate dead code if the session exists.

I'm not well versed in Ruby code - Apologies if I've done something crazy!

### DIFF
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -147,6 +147,12 @@ module Tmuxinator
       yaml["tmux_command"] || "tmux"
     end
 
+    def tmux_has_session?(name)
+      sessions = `#{tmux_command} ls`
+
+      ! sessions.match("^#{name}:");
+    end
+
     def socket
       if socket_path
         " -S #{socket_path}"
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -150,7 +150,7 @@ module Tmuxinator
     def tmux_has_session?(name)
       sessions = `#{tmux_command} ls`
 
-      !sessions.match("^#{name}:")
+      !!sessions.match("^#{name}:")
     end
 
     def socket
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -150,7 +150,7 @@ module Tmuxinator
     def tmux_has_session?(name)
       sessions = `#{tmux_command} ls`
 
-      ! sessions.match("^#{name}:");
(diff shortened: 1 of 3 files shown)


Richard-Walton:
Fix for: https://github.com/tmuxinator/tmuxinator/issues/402

Also, improved template rendering to eliminate dead code if the session exists.

I'm not well versed in Ruby code - Apologies if I've done something crazy!

lib/tmuxinator/project.rb
```diff
@@ -147,6 +147,12 @@ def tmux_command
       yaml["tmux_command"] || "tmux"
     end
 
+    def tmux_has_session?(name)
+      sessions = `#{tmux_command} ls`
+
+      !sessions.match("^#{name}:")
```
J3RN: I'm a bit confused as to why there's a `!` here.

The way I read it, if the session exists, `session.match("^#{name}:")` should return a match. `not`ing a match would return `false`, which is returned from the method. Shouldn't the method return `true` if the session exists?

If you were trying to make a boolean from the match object, you can do `!!`. For example:
```ruby
!! "foo" #=> true
```

Richard-Walton: Err..., yeah that's just poor logic.  My bad - I was rushing to get this pushed so I could get back to work.

Fixed in next commit.

Richard-Walton:
Anything else need changing?

adamstrickland:
@Richard-Walton thanks for asking...  I would like to see some tests.  I think two tests for `Project#tmux_has_session?` are called for, one for the positive case, one for the negative.  Something along the lines of:

```ruby
  describe "#tmux_has_session?" do
    let(:reply) { "foo: 1 window (created blah blah blah) [500x500] (detached)\nbar: 1 window (created blah blah blah) [500x500] (detached)" }

    before do
      expect(Kernel).to receive(:system).with("tmux ls"){ reply }
    end

    it "should be true when 'tmux ls` has the named session" do
      expect(subject.tmux_has_session?("foo")).to be_true
    end

    it "should be false when 'tmux ls` does not have the named session" do
      expect(subject.tmux_has_session?("quux")).not_to be_true
    end
  end
```

Richard-Walton:
@adamstrickland Ok, added tests.  Again, I'm not a ruby dev so this code may be crazy. 

Related note: The rubocop line length constraint is difficult to work with.  This some of the multiline strings in the tests.

Issue #406, opened 2016-04-14
Strange behaviour when panes defined as multi-element dictionaries


podung:
There appears to be a bug in the handling of a pane element that contains a multi-element dictionary.  I found it when working on a PR for #266.

```yaml
# This works: A pane with 2 elements in an array, each being a single-element dictionary
panes:
  - pane1: echo "first pane"
  - pane2: echo "second pane"
# results in the expected 2 panes, each running its respective echo command.
```

```yaml
# This doesn't: An array that contains a multi-element dictionary
panes:
  - pane1: echo "first pane"
    pane2: echo "second pane"
# bizarrely creates three panes
# All the commands run in the first pane, the next two panes are empty panes
```

The options I see are:
1. Take the first element and ignore the rest (this is consistent with how multi-element Windows yaml is handled (`value = window_yaml.values.first` in window.rb:18)
2. Make each element its own pane
3. Print out some sort of invalid format error

I know its a corner case that's pretty minor, but I'm trying clean up the surrounding code a bit while I address #266.

Any thoughts?  I'm open to any suggestions.

podung:
For anyone interested, the source of this bug is the pane_yml hash gets mapped and a new Pane is created for each entry.  The problem is the same index is passed to each of these objects:

```ruby
# window.rb

def build_panes(panes_yml)
  Array(panes_yml).map.with_index do |pane_yml, index|
    if pane_yml.is_a?(Hash)
      pane_yml.map do |_name, commands|
        Tmuxinator::Pane.new(index, project, self, *commands)
      end
    else
      Tmuxinator::Pane.new(index, project, self, pane_yml)
    end
  end.flatten
end
```

In the sample case above this creates 2 Pane objects each with an index of 0.  When the template.erb outputs the pane commands all of the commands target get run against pane 0.  

Additionally this breaks the logic for determining `pane.last?`, resulting in 1 pane that runs all the commands and two additional empty panes.

podung:
I guess each hash element could also be interpreted as a command to run in the pane.  But its kind of weird to name your commands with hash keys.

podung:
Closed by #418

PR #418, opened 2016-05-17
266 initialize pane commands with yaml array

### LINKED ISSUE #266: Pane with multiple commands requires name
If I create a window with multiple panes and then I want multiple commands to be run in one of the panes I need to assign a (useless) name.

This doesn't work:
```
name: test
root: ~/

windows:
  - editor:
      layout: main-vertical
      panes:
        -
          - cd tmp
          - vim testfile
        - guard
```

But this does:

```
name: test
root: ~/

windows:
  - editor:
      layout: main-vertical
      panes:
        - dumb:
          - cd tmp
          - vim testfile
        - guard
```

Though both are valid yaml.

podung:
Resolves #266 and #406 

It was easier to test my behavior using actual results not just stubbing calls to `Tmuxinator::Pane.new`.  I added a custom rspec matcher to keep the spec looking clean and spiffy.

I also removed an unused code path related to `Pane#mutliple_commands?`.  Because the pane's command arg is splatted there is always an array and it is sufficient to iterate it regardless of how many commands are contained within.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unlreased
+
+### New Features
+- Allow mulitple panes to be defined using yaml hash or array #266, #406
+
 ## 0.8.1
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unlreased
+## Unreleased
 
 ### New Features
 - Allow mulitple panes to be defined using yaml hash or array #266, #406
--- a/lib/tmuxinator/window.rb
+++ b/lib/tmuxinator/window.rb
@@ -34,13 +34,16 @@ module Tmuxinator
 
     def build_panes(panes_yml)
       Array(panes_yml).map.with_index do |pane_yml, index|
-        if pane_yml.is_a?(Hash)
-          pane_yml.map do |_name, commands|
-            Tmuxinator::Pane.new(index, project, self, *commands)
-          end
-        else
-          Tmuxinator::Pane.new(index, project, self, pane_yml)
-        end
+        commands =  case pane_yml
+                    when Hash
+                      pane_yml.values.first
+                    when Array
+                      pane_yml
+                    else
+                      pane_yml
+                    end
+
+        Tmuxinator::Pane.new(index, project, self, *commands)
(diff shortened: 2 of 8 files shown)


J3RN:
This PR looks good! ✨ I have a few nitpicks, though, so I'm going to go through and comment on the appropriate lines.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unlreased
```
J3RN: I think you meant "Unreleased" 😄

lib/tmuxinator/window.rb
```diff
@@ -34,13 +34,16 @@ def initialize(window_yaml, index, project)
...
-        if pane_yml.is_a?(Hash)
-          pane_yml.map do |_name, commands|
-            Tmuxinator::Pane.new(index, project, self, *commands)
-          end
-        else
-          Tmuxinator::Pane.new(index, project, self, pane_yml)
-        end
+        commands =  case pane_yml
+                    when Hash
+                      pane_yml.values.first
```
J3RN: Is there a situation in which there would be multiple elements in this Hash? As in, might the hash contain something like this?
```ruby
{ "one" => ["cd tmp", "vim"], "two" => ["cd log", "tail -f development.log"] }
```

If so, I think the easy fix is just changing this line to `pane_yml.values`

CHANGELOG.md
podung: Yep.  Thanks.  I'll fix this.

lib/tmuxinator/window.rb
podung: Yeah I wrestled with the right behavior here.  If you could check out #406 and give me your thoughts.

If you think about the yaml structure, its not clear what exactly multiple hash elements would represent.  I gave a few options in that issue above.

J3RN: Alright, now that I've read that issue and understand the premise a little bit more, I think your approach here with using `first` is right. I'm going to pull this down and play with it a little just to make sure, then I'll come back and give it a :+1:

podung:
Cool. I can fix the misspelling tonight

J3RN:
Alright, it seems to be working. LGTM :+1:

Issue #401, opened 2016-04-04
'mux' symlink incorrectly installed on Ruby 2.3.0 (RubyGems issue)


e2:
FYI: https://github.com/rubygems/rubygems/issues/1577

ohcibi:
I have ruby 2.2.3 and the same issue.

e2:
The workaround is to just copy the file until the symlinking is fixed.

It's better to have a gem with duplicate files (no matter how "wrong" that feels) rather than a non-working gem.

J3RN:
Something I have been contemplating is removing the `mux` file altogether, and just suggesting that people alias `mux` to `tmuxinator`.

ethagnawl:
@J3RN @ncoop Shall we close this issue?

J3RN:
Since this is a RubyGems issue and the issue has been appropriately opened on that repo, I will close this issue.

ncoop:
Out of curiosity, how have other gem maintainers dealt with this?

J3RN:
@ncoop Hard to say, but we're the only ones who have referenced the RubyGems issue.

PR #419, opened 2016-05-20
Fix symlink to mux

timeemit:
Love `tmuxinator` but hate spelling it out on the command line.

The aim of this PR serves to address #401 by resolving the path from the root directory of the project.

### DIFF
--- a/bin/mux
+++ b/bin/mux
@@ -1,1 +1,1 @@
-tmuxinator
+bin/tmuxinator


timeemit:
Ah, just read through [the actual problem](https://rubygems/issues/1577).

Happy to write a copy of `tmuxinator` to _bin/mux_ like @e2 suggests in #401 if that's the consensus.

PR #97, opened 2013-02-11
Synchronize panes support.

jessegonzalez:
I synchronize-panes all the time, and so may others.

PR #423, opened 2016-06-02
Add support for window synchronization.

fabianschwarzfritz:
Implements #97 after refactoring. **Thanks @jessegonzales for the idea and
first implementation!**. 🙇  
As the PR  #97 is quite old, here is the same thing, re-implemented on a more recent codebase. Blame me, if doing that is not welcome 😢 

Unfortunately I am not sure about the mocks in the  `describe "#tmux_synchronize_panes"` section. Maybe someone can have a look? Feedback welcome 🍻

### DIFF
--- a/spec/lib/tmuxinator/window_spec.rb
+++ b/spec/lib/tmuxinator/window_spec.rb
@@ -261,17 +261,16 @@ describe Tmuxinator::Window do
     end
 
     let(:tmux_part) { project.tmux }
-    let(:window_option_set) { "set-window-option" }
+    let(:window_option_set_part) { "set-window-option" }
     let(:target_part) { "-t #{window.tmux_window_target}" }
-    let(:synchronize_panes) { "synchronize-panes" }
-    let(:tmux_part) { project.tmux }
+    let(:synchronize_panes_part) { "synchronize-panes" }
 
     context "synchronization enabled" do
       let(:synchronize) { true }
       let(:enabled) { "on" }
 
       let(:full_command) do
-        "#{tmux_part} #{window_option_set} #{target_part} #{synchronize_panes} #{enabled}"
+        "#{tmux_part} #{window_option_set_part} #{target_part} #{synchronize_panes_part} #{enabled}"
       end
 
       it "should set the synchronize-panes window option on" do
--- a/spec/lib/tmuxinator/window_spec.rb
+++ b/spec/lib/tmuxinator/window_spec.rb
@@ -271,10 +271,7 @@ describe Tmuxinator::Window do
       let(:enabled) { "on" }
 
       let(:full_command) do
-        command = <<END_COMMAND
-#{tmux_part} #{window_option_set} #{target_part} #{synchronize_panes} #{enabled}
-END_COMMAND
-        command.chomp("\n")
+        "#{tmux_part} #{window_option_set} #{target_part} #{synchronize_panes} #{enabled}"
       end
 
       it "should set the synchronize-panes window option on" do
--- a/spec/lib/tmuxinator/window_spec.rb
+++ b/spec/lib/tmuxinator/window_spec.rb
@@ -251,7 +251,7 @@ describe Tmuxinator::Window do
     context "synchronization disabled" do
       let(:synchronize) { false }
 
-      it "should have synchronization enabled" do
+      it "should have synchronization disabled" do
         expect(window.synchronize).to be false
       end
     end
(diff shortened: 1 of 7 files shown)


spec/lib/tmuxinator/window_spec.rb
```diff
@@ -208,6 +210,69 @@
...
+    context "synchronization not specified" do
+      it "should have synchronization disabled" do
+        expect(window.synchronize).to be false
+      end
+    end
+  end
+
+  describe "#tmux_synchronize_panes" do
+    let(:project) { double(:project) }
+    let(:window) { Tmuxinator::Window.new(yaml, 0, project) }
```
fabianschwarzfritz: Honestly, I am not sure about this stuff.

spec/lib/tmuxinator/window_spec.rb
```diff
@@ -208,6 +210,69 @@
...
+    let(:synchronize_panes) { "synchronize-panes" }
+    let(:tmux_part) { project.tmux }
+
+    context "synchronization enabled" do
+      let(:synchronize) { true }
+      let(:enabled) { "on" }
+
+      let(:full_command) do
+        command = <<END_COMMAND
+#{tmux_part} #{window_option_set} #{target_part} #{synchronize_panes} #{enabled}
```
fabianschwarzfritz: And this is because the code linter doesn't like more than 80 characters. I'm not a ruby expert. Any suggestions for a better way to implement that line?

adamstrickland: Add the file path to this array: https://github.com/tmuxinator/tmuxinator/blob/3ab8b98dda9611803839cb6663b3e0fd28ad31a4/.rubocop.yml#L509-L510

fabianschwarzfritz:
Like that?

adamstrickland:
yes

spec/lib/tmuxinator/window_spec.rb
adamstrickland: Not sure what the question is here...

fabianschwarzfritz:
The travis build failed - https://travis-ci.org/tmuxinator/tmuxinator/jobs/134830222 . Don't think that's an issue of the coding, but `gem install bundler` fails. 
Anyway I am in vacation for 2 weeks 🐚 ☀️ 🌊 , and I'll look into this after I'm back, okay?

spec/lib/tmuxinator/window_spec.rb
fabianschwarzfritz: Sorry, I admit that was quite unspecific. 
Nearly the same doubles/stubs are now defined in the description of `tmux_new_window_command` and `tmux_synchronize_panes` . That's an ugly duplication I probably should get rid of, or is that not an issue? Should I invest some more time here?

adamstrickland: Ah....  I think what you want here is an RSpec [shared context](https://relishapp.com/rspec/rspec-core/v/3-4/docs/example-groups/shared-context) to make that setup more DRY.

```ruby
describe Tmuxinator::Window do
  shared_context "new window setup" do
    let(:project) { double(:project) }
    let(:window) { Tmuxinator::Window.new(yaml, 0, project) }
    # etc
  end

  describe "#tmux_synchronize_panes" do
    include_context "new window setup"

    it "blah blah" do
    end
  end

  describe "#tmux_new_window_command" do
    include_context "new window setup"

    it "foo bar bazz" do
    end
  end
end
```

Probably a good idea, yes.

fabianschwarzfritz: Thanks for the suggestion and your quick feedback, @adamstrickland . I added such a shared context with 5837d3e .

adamstrickland:
👍 

@J3RN @podung your thoughts?

spec/lib/tmuxinator/window_spec.rb
```diff
@@ -208,23 +229,54 @@
...
+      it "should have synchronization enabled" do
+        expect(window.synchronize).to be true
+      end
     end
 
-    let(:tmux_part) { project.tmux }
+    context "synchronization disabled" do
+      let(:synchronize) { false }
+
+      it "should have synchronization enabled" do
```
podung: minor: 
I believe this should read `it "should have synchronization disabled" do`

podung:
This looks great.  Thanks @fabianschwarzfritz 

Other than the minor typo in the spec it block, my only ask is if you could add the same logic to the wemux_template.erb.  Its on my wish list to get these templates as identical as possible and then abstract out everything that is common.

J3RN:
LGTM! :shipit:

podung:
Thanks so much @fabianschwarzfritz.  This is great!

PR #428, opened 2016-06-08
Fix for windows without panes

### LINKED ISSUE #328: 'mux start' not working: undefined method 'each' for nil:NilClass (NoMethodError)
I installed tmuxinator using 'sudo gem install tmuxinator', created a project and then attempted to start it. I get the following failure:

    $ mux start foo
      (erubis:41:in `block in render': undefined method `each' for nil:NilClass (NoMethodError)
            from (erubis:34:in `each'
            from (erubis:34:in `render'
            from /var/lib/gems/1.9.1/gems/erubis-2.7.0/lib/erubis/evaluator.rb:65:in `eval'
            from /var/lib/gems/1.9.1/gems/erubis-2.7.0/lib/erubis/evaluator.rb:65:in `result'
            from /var/lib/gems/1.9.1/gems/tmuxinator-0.6.11/lib/tmuxinator/project.rb:17:in `render'
            from /var/lib/gems/1.9.1/gems/tmuxinator-0.6.11/lib/tmuxinator/cli.rb:77:in `start'
            from /var/lib/gems/1.9.1/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
            from /var/lib/gems/1.9.1/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
            from /var/lib/gems/1.9.1/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
            from /var/lib/gems/1.9.1/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
            from /var/lib/gems/1.9.1/gems/tmuxinator-0.6.11/bin/mux:18:in `<top (required)>'
            from /usr/local/bin/mux:23:in `load'
            from /usr/local/bin/mux:23:in `<main>'
    
    $ tmuxinator version
    tmuxinator 0.6.11
    $ tmuxinator doctor
    Checking if tmux is installed ==> Yes
    Checking if $EDITOR is set ==> Yes
    Checking if $SHELL is set ==> Yes
    $ tmux -V
    tmux 2.0

This is on Linux.
Please let me know which other information I can provide. 
I did not find any similar issues, so there must be something wrong in my installation.

J3RN:
Fixes #328

The simple idea is that the `commands` instance variable of the `Window` class would not be initialized under certain conditions. This change ensures that `commands` will always be initialized, defaulting to an empty array.

Kudos to @fikovnik for the fix.

### DIFF
--- a/lib/tmuxinator/window.rb
+++ b/lib/tmuxinator/window.rb
@@ -27,9 +27,9 @@ module Tmuxinator
                 end
 
         @panes = build_panes(value["panes"])
-      else
-        @commands = build_commands(tmux_window_command_prefix, value)
       end
+
+      @commands = build_commands(tmux_window_command_prefix, value)
     end
 
     def build_panes(panes_yml)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Allow mulitple panes to be defined using yaml hash or array #266, #406
 - Add `startup_pane` #380
 
+### Bugfixes
+- Will no longer crash when no panes are specified in a window
+
 ## 0.8.1
 ### Bugfixes
 
--- a/spec/lib/tmuxinator/window_spec.rb
+++ b/spec/lib/tmuxinator/window_spec.rb
@@ -188,6 +188,16 @@ describe Tmuxinator::Window do
         expect(window.commands).to be_empty
       end
     end
+
+    context "command is a hash" do
+      before do
+        yaml["editor"] = { "layout" => "main-horizontal", "panes" => [nil] }
+      end
+
+      it "returns an empty array" do
+        expect(window.commands).to be_empty
+      end
+    end
   end
 
   describe "#name_options" do


lib/tmuxinator/window.rb
```diff
@@ -10,6 +10,7 @@ def initialize(window_yaml, index, project)
               end
       @root = nil
       @panes = []
+      @commands = []
```
adamstrickland: Should we add a test for this, that verifies a valid `commands` instance, given an empty `panes` in the yaml?

J3RN: Good thought -- I'll add that.

J3RN:
@adamstrickland After looking a little closer, it looks like the `commands` variable is populated by the `build_commands` method. This method properly handles all cases, so I popped it out of the `else` case it was living in, and tested for the case it was protected against.

On a philosophical note, the `build_commands` method is only used inside the `Window` class, so it should probably be a private method, and it's functionality should be tested by way of the `commands` method (an `attr_reader` of the instance var). In any case, that's why I didn't add tests for the `commands` method --- it would just be duplicating the tests of the `build_commands` method.

adamstrickland:
Cool.  LGTM

J3RN:
@ethagnawl @podung Thoughts?

I'll squash before we merge, just wanted to make sure no new changes are needed.

podung:
I'll finish looking this over in the morning. Just want to play with some yaml files before giving it a thumbs up

podung:
I'm finally looking at this now.  Will merge after I'm done.

podung:
@J3RN I can't recreate the issue people have been complaining about.

In looking at the code, I think that the template iterates over the panes collection if they exist and only if they do not iterates over the commands collection.

```
    <% unless window.panes? %>
      ... 
      <% window.commands.each do |command| %>
  <%= command %>
      <% end %>
    <% else %>
      <% window.panes.each do |pane| %>
      ...
      <% end %>
   <% end %>
```

So even if `@commands` is not initialized, I don't see how that would cause it to blow up.  Have you been able to recreate the issue?

podung:
Ok cool.  I can recreate now.

Looking through the code the duplication in dealing with commands in `Tmuxinator::Window` and `Tmuxinator::Pane` jumped out at me:

```ruby
#Tmuxinator::Window
def tmux_pre_window_command
  return unless project.pre_window

  "#{project.tmux} send-keys -t #{tmux_window_target} #{project.pre_window.shellescape} C-m"
end

#Tmuxinator::Pane
def tmux_pre_window_command
  return unless project.pre_window

  "#{project.tmux} send-keys -t #{tmux_window_and_pane_target} #{project.pre_window.shellescape} C-m"
end
```

Also, both have to know about command(s) to be run in the pane/window.  The duplication is really obvious when you look at `lib/assets/template.erb`:
```ruby
# Window "<%= window.name %>"
unless window.panes?
  if window.project.pre_window
    window.tmux_pre_window_command
  end
  window.commands.each do |command|
    command
  end
else
  window.panes.each do |pane|
    if pane.project.pre_window
      pane.tmux_pre_window_command
    end
    pane.commands.each do |command|
      pane.tmux_main_command(command)
    end
  end
end
```

We shouldn't have the concept of a window without any panes.  In that case, we should just create a backing pane object and pass any window commands down to it.

Your fix works for the issue at hand though and it looks good to me.  Did you want to squash still or should I go ahead and merge?

J3RN:
I still want to squash. I'll do that and merge it in the morning.

PR #436, opened 2016-07-08
Replace mux symlink with alias

ncoop:
Simple solution to a complicated problem. (#401)
Aliasing occurs when the user sources the completion file in their *shrc.
Not sure how to write a test spec for this.
Also, not sure why mux.fish exists.


ethagnawl:
IMO, the completion hack is a good stopgap. 

However, I'd vote that bin/mux stays or at least gets tabled and has a new issue is opened about re-introducing it, so that if/when Rubygems gets this issue sorted out, `mux` will 1) work out-of-the-box again and 2) work for people who don't want to/can't use shell completions.

J3RN:
I strongly encourage this change :+1:

@ncoop Would you mind updating the `CHANGELOG`?

ncoop:
@J3RN OK, done.
@ethagnawl Should the symlink be left alone then?

ethagnawl:
@ncoop If it doesn't cause any problems, I don't see why not. I don't feel too strongly about it, though - I'd just like to see that option working as it was again at some point. Does anyone else have thoughts on this?

J3RN:
@ncoop Side note -- the file modes changed for completions. Was that change intentional?

@ethagnawl I'm concerned that RubyGems will probably never solve this issue since they don't seem to have given it much thought thus far. Currently, the 'mux' file does produce an ugly warning when installing tmuxinator and tells users to run a command that does not help, which is a little concerning. It feels to me that the 'mux' file has always been a placeholder for an alias -- that or that the executable should actually be named 'mux'. Then again, I'm not going to enforce these ideas :smiley:

ncoop:
@J3RN Yes. The completions get read by `source`, so there is no need for them to be executable.

Well, if the symlink is there, we need to do something to confirm it's not valid before aliasing, yeah?

J3RN:
Oh &mdash; nice call on those permissions. I should have looked at that more closely :smiley:

I am in favor of removing the symlink.

ethagnawl:
@ncoop Sorry - I'd forgotten that it was causing an error on install. I recant my earlier suggestion and vote to remove it, until Rubygems fixes the bug.

J3RN:
What is the current status of this?

J3RN:
@ncoop It looks like you'll need to rebase. Sorry about that.

@ethagnawl @adamstrickland Are either of you opposed to merging this?

adamstrickland:
LGTM  👍

ncoop:
Rebased.

PR #446, opened 2016-08-03
Only enable Synchronize panes after the commands have run

DuncSmith:
Further to PR https://github.com/tmuxinator/tmuxinator/pull/423, if synchronization is enabled. As each pane is created, existing panes also receive each command string.

For example
```
windows:
  - editor:
      layout: main-vertical
      synchronize: true
      panes:
        - echo 'pane 1'
        - echo 'pane 2'
        - echo 'pane 3'
        - echo 'pane 4'
```
Results in
image

I'm not sure if this is desired behaviour. However in my use case (ssh'ing to multiple servers) I found it could potentially cause issues.

Addressed by moving the call to sync panes to after the pane creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 - Allow mulitple panes to be defined using yaml hash or array #266, #406
 - Add `startup_pane` #380
 - Add synchronizations panes support #97
-- Add `before` and `after` options to synchronisation functionality
+- Add `before` and `after` options to synchronization functionality
+- Add deprecation warning if `synchronize: true` or `before` is used
 
 ### Bugfixes
 - Supress `tmux ls` non-zero exit status/message when no sessions exist (#414)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Will no longer crash when no panes are specified in a window
 - Locking activesupport at < 5.0.0 to prevent broken builds on Ruby < 2.2.3
 - Fixed whitespace issues in help
+- Initial commands no longer sent to multiple panes when synchronize panes enabled
 
 ## 0.8.1
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,13 +11,13 @@
 - Allow mulitple panes to be defined using yaml hash or array #266, #406
 - Add `startup_pane` #380
 - Add synchronizations panes support #97
+- Add `before` and `after` options to synchronisation functionality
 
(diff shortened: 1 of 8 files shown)


adamstrickland:
Wow, I had no idea that this is what happened (obviously I don't personally use that option much).

I assume that the desired output here is to have `echo 'pane 1'` run in - and only in - pane 1, `echo 'pane 2'` only run in pane 2, etc., correct?  If so, that seems to me to be the intuitive behavior of the feature in general.

However, I'm a little concerned that this would break things for users that may already depend on the existing behavior.  Although we've talked about adding some analytics so we can track usage on features like that, we haven't implemented it yet, so we don't really know...  I'd like to see if we can get a sense of how others use this functionality before merging.  Thoughts?

DuncSmith:
Yes, that's how I perceived the feature should work. I think the existing behaviour has potential to cause strange issues, I wouldn't expect all the commands to be sent to the first pane for example. I think the correct way to implement that would be something more like:

```
windows:
  - editor:
      layout: main-vertical
      synchronize: true
      panes:
        - pane_1:
          - echo 'pane 1'
          - echo 'pane 2'
          - echo 'pane 3'
          - echo 'pane 4'
        - pane_2:
          - echo 'pane 1'
          - echo 'pane 2'
          - echo 'pane 3'
          - echo 'pane 4'
```

adamstrickland:
@DuncSmith I started a conversation on this over in our Gitter channel, in case you'd like to chime in: https://gitter.im/tmuxinator/tmuxinator

ethagnawl:
@DuncSmith Looking good! 

My one remaining question is: should the user see a deprecation warning if they use `synchronize: true` or `synchronize: 'before'`?

adamstrickland:
IMO, yes, we should have a deprecation notice.  Also needs to be added to the changelog, etc.

DuncSmith:
@adamstrickland @ethagnawl Apologies for delay, updated and rebased.

adamstrickland:
@DuncSmith np.  thanks for the help.  👍

ethagnawl:
@DuncSmith Ditto! 

I just pulled your branch, did some manual QA and it looks great. 

My (superficial) feedback would be:
- change "ran" to "run" in `SYNC_DEP_MSG` 
- change "0.8.2" to "a future release" in `SYNC_DEP_MSG`

adamstrickland:
Damn, I merged too fast...  I'll PR your suggested changes @ethagnawl since I did the merge.  Now I remember why no one trusts me :-)

PR #450, opened 2016-08-18
tweaks synchronization deprecation message

adamstrickland:


ethagnawl:
@adamstrickland What would you think about re-introducing the last sentence of @DuncSmith's deprecation warning (or something similar), which explicitly tells users how to opt into the `after` behavior? (e.g. "To use this behaviour now use the 'synchronize: after' option.")

adamstrickland:
@ethagnawl done

Issue #385, opened 2016-03-11
Better validation/error handling when Tmux is not installed


ethagnawl:
For instance, calling `mux start foo` (when Tmux has not been installed) doesn't fail until execution reaches Project#render. I think ideally, the first thing Config::validate would do is raise an exception if Tmux is not available.

PR #452, opened 2016-08-29
385 introduce tmux validator [WIP]

ethagnawl:
Addresses #385.

A few items of note:
- "validation" currently takes place in `Cli#start`, but that's definitely up for debate. Personally, I'd run the "validation" inside the Tmuxinator module definition. However, it's feasible that someone would want to use Tmuxinator to delete existing projects, run `tmuxinator implode`, etc. after having uninstalled tmux.
- TmuxValidator currently raises a `RuntimeError` when tmux is unavailable. I'd prefer to raise a custom error like, `TmuxIsNotAvailable`, but I don't believe that's being done anywhere else in the project, so I don't want to break formation. It might also make sense to use `Util#exit!` to avoid dumping a stack trace on unassuming users.

### DIFF
--- a/lib/tmuxinator.rb
+++ b/lib/tmuxinator.rb
@@ -15,7 +15,8 @@ require "tmuxinator/window"
 require "tmuxinator/version"
 require "tmuxinator/tmux_validator"
 
-module Tmuxinator; end
+module Tmuxinator
+end
 
 class Object
   def blank?
--- a/lib/tmuxinator.rb
+++ b/lib/tmuxinator.rb
@@ -13,9 +13,9 @@ require "tmuxinator/pane"
 require "tmuxinator/project"
 require "tmuxinator/window"
 require "tmuxinator/version"
+require "tmuxinator/tmux_validator"
 
-module Tmuxinator
-end
+module Tmuxinator; end
 
 class Object
   def blank?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 - Add synchronizations panes support #97
 - Add `before` and `after` options to synchronization functionality
 - Add deprecation warning if `synchronize: true` or `before` is used
+- Add TmuxValidator, which currently only validates that tmux has been installed
+  and is available on $PATH. In the future, this module could also be used to
+  validate minimum tmux version, etc. #385
 
 ### Bugfixes
 - Supress `tmux ls` non-zero exit status/message when no sessions exist (#414)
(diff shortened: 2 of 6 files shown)


lib/tmuxinator.rb
```diff
@@ -13,9 +13,9 @@
 require "tmuxinator/project"
 require "tmuxinator/window"
 require "tmuxinator/version"
+require "tmuxinator/tmux_validator"
 
-module Tmuxinator
-end
+module Tmuxinator; end
```
J3RN: While this isn't really relevant to this PR, per se, it turns out that this line is actually unnecessary.

ethagnawl: Good point. I'll back that out and it can be formatted/removed elsewhere, if people feel strongly about it.

ethagnawl:
> Tmuxinator::Config actually contains several not config-related methods, such as installed?, editor?, and shell?. Should we open an issue to move these to a validator of sorts?

Agreed. I'll open an issue about that.

Did you have any thoughts about the placement of the call to `Tmuxinator::TmuxValidator::validate!`, @J3RN? That's the only aspect of this feature I have mixed feelings about.

J3RN:
@ethagnawl I think `validate!`'s current position is fine, given our current codebase.

An idea I was toying with was running "doctor" silently before `start` (or even others), which would produce warnings if any of the elements aren't met. The key difference between that implementation and the one here is that a warning wouldn't stop execution, but this implementation of using `raise` does (which I honestly prefer). Using raise in "doctor" would be counterproductive, since the current implementation of "doctor" gives the user a checklist of things to fix.

What were you thinking about for an alternate implementation?

ethagnawl:
> An idea I was toying with was running "doctor" silently before start ...

Yeah, that's a great idea. There's definitely something to be fleshed out with
respect to the Doctor and Validator concepts. Maybe we create one module, 
which does all of the above and accepts an argument which determines 
whether it prints warning messages or raises exceptions? I'm open to 
scrapping this feature if it makes sense to nail down that implementation, 
rather than ship something that's half-baked. If not, we should definitely 
continue that conversation in #457.

> What were you thinking about for an alternate implementation?

I was thinking the validation could take place somewhere inside Tmuxinator.rb.
In my opinion, execution should halt if Tmux isn't available - that prevents us
from having to shoehorn validation into every possible entrypoint (... and now
that I just said that, I realize this current implementation doesn't cover
every scenario; e.g. `Cli#stop`). However, if that's too heavy handed, then
`Cli#start` is an obvious candidate for a scenario where validation should
happen. (Per my revelation, we'll need to map out each entrypoint which calls
out to tmux if we want to go with the individual entrypoint approach.)

PR #453, opened 2016-08-29
Fix wemux support

athal7:


athal7:
@J3RN sounds like there's some desire for this fix, do you have a chance to take a look?

adamstrickland:
An (unfortunately) long-neglected PR...  @J3RN, you had self-assigned this; is it ready to bring into master?

PR #454, opened 2016-09-08
Add name flag to `tmuxinator start` documentaion

jackturnbull:
The way tmuxinator currently works is to expect the name (or alias) of the start command to be passed in underneath the `-n` flag. The documentation has been updated to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,11 +260,11 @@ root: ~/<%= @settings["workspace"] %>
 This will fire up tmux with all the tabs and panes you configured.
 
 ```
-tmuxinator start [project] [alias]
+tmuxinator start [project] -n [name]
 ```
 
-If you use the optional `[alias]` argument, it will start a new tmux session
-with the custom alias name provided.  This is to enable reuse of a project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,11 +260,11 @@ root: ~/<%= @settings["workspace"] %>
...
 
 ```
-tmuxinator start [project] [alias]
+tmuxinator start [project] -n [name]
 ```
 
-If you use the optional `[alias]` argument, it will start a new tmux session
-with the custom alias name provided.  This is to enable reuse of a project
+If you use the optional `[name]` argument, it will start a new tmux session
+with the custom name name provided.  This is to enable reuse of a project
```
J3RN: This change looks good! My only commentary is to just say "custom name provided" instead of "custom name name provided".

jackturnbull: Embarrassing, sorry! Fixed.

J3RN:
Alright, this looks good to me! Thanks for your help, @jackturnbull!

Issue #458, opened 2016-09-22
Remove Object#blank? monkey patch


ethagnawl:
`Object#blank?` is currently being monkey-patched in Tmuxinator.rb. 

However, the project has a dependency on Active Support (see #432), which is where the source of this monkey patch originates. So, as long as the project maintains that dependency, the monkey patch in Tmuxinator.rb isn't necessary.

Soliah:
ActiveSupport was removed in #199 as it seemed unnecessary. That PR also added the monkey patch I believe. If factory_girl was pulling it in as a dep it should've been limited to a development dependency.

J3RN:
Whoa, I just realized that there's gems in our Gemfile. We should probably remove those.

ethagnawl:
Closing this in favor of #459.

J3RN:
To be fair, I think we should still remove the monkey patch. A quick `ag` says that `blank?` is only used twice. I'd have to do a bit of digging, but I suspect it probably wouldn't be too hard to change those two instances.

Soliah:
@ethagnawl Just thought I'd add that ActiveSupport isn't a runtime dep, so though factory_girl is included via bundler, it's not required during install of the gem.

PR #461, opened 2016-09-23
introduce and use noroot fixture/factory in project_spec

ethagnawl:
I noticed this omission while looking into #458.

- introduces noroot fixture
- introduces factory for noroot fixture
- uses noroot factory (instead of noname factory) in "without root" test


J3RN:
The more tests the better! :+1:

Issue #459, opened 2016-09-23
Remove Gems from Gemfile


J3RN:
I can't find the documentation on this, but I believe it's an antipattern to have dependencies listed in a gem's Gemfile. As meager evidence of this, I included a screenshot of the default Gemfile when creating a new gem:
![screen shot 2016-09-23 at 10 37 54 am](https://cloud.githubusercontent.com/assets/2058614/18789759/2447da12-817a-11e6-9997-66ea7dde307d.png)

If I'm right about this, we should remove the gems listed in our Gemfile.

adamstrickland:
I think the original intent (just speculating here, based on the gems enumerated) was to keep development/testing/tooling dependencies in the the Gemfile, and keep the runtime dependencies in the gemspec.  That being said, there's no reason why we couldn't move the Gemfile deps to the gemspec as `#add_development_dependency` line items.

ethagnawl:
@J3RN, I think you are right. I was actually reading [this](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) blog post just now, which makes that case.

Soliah:
@adamstrickland Looks like it was me merging #311. Originally I did have the development deps in the gemspec but looks like I must of missed the change in the that PR.

Issue #432, opened 2016-07-03
ActiveSupport 5.0 breaks build for Ruby < 2.2.3


ethagnawl:
This dependency appears to be coming from factory_girl. 

I took a quick pass at addressing this and at appears that locking ActiveSupport at < 5.0.0 in Tmuxinator's Gemfile solves the problem ([gist](https://gist.github.com/ethagnawl/baa517a8537424532815e75bea436037)).

adamstrickland:
So, do we want to 🚢  this and then have https://github.com/tmuxinator/tmuxinator/pull/416 pull latest before merging in the PR?

adamstrickland:
As an aside, why are we putting this in the Dockerfile and not in the Gemfile itself?

ethagnawl:
> So, do we want to :ship: this and then have #416 pull latest before merging in the PR?

That sounds good by me.

> As an aside, why are we putting this in the Dockerfile and not in the Gemfile itself?

That was just me reproducing the broken build. The change in the [PR](https://github.com/tmuxinator/tmuxinator/pull/433/files) was made to the Gemfile.

adamstrickland:
Ah cool 👍

PR #464, opened 2016-10-02
Test tmux versions 2.2 and 2.3

seanmalloy:


tmuxinator.gemspec
```diff
@@ -42,5 +42,14 @@ Gem::Specification.new do |s|
...
 
   s.add_development_dependency "bundler", "~> 1.3"
+  s.add_development_dependency "rake", "~> 10.4"
+  s.add_development_dependency "rspec", "~> 3.3"
+  s.add_development_dependency "simplecov", "~> 0.11.0"
+  s.add_development_dependency "coveralls", "~> 0.7"
+  s.add_development_dependency "awesome_print", "~> 1.2"
+  s.add_development_dependency "pry", "~> 0.10"
+  s.add_development_dependency "factory_girl", "~> 4.5"
+  s.add_development_dependency "activesupport", "< 5.0.0"  # Please see issue #432
```
J3RN: Hey there! I'm loving this pull request. My only comment is whether this particular line, adding activesupport as a dependency (granted, only a development dependency) is actually needed.

I'll do a bit of digging to see whether it's needed, and will get back to you.

Reference: #432, #458
Fixes #459

seanmalloy: I included the activesupport gem because it was in the original Gemfile. I can remove it from the gemspec if it is not required.

J3RN: From my testing, it looks like it's not needed. Would you mind removing it?

Also, please throw a not in the CHANGELOG.md file. Thanks!

seanmalloy: Removing the activesupport gem breaks the Travis CI tests for Ruby 1.9.3. I updated the CHANGELOG.

Gemfile
```diff
@@ -3,12 +3,4 @@ source "https://rubygems.org"
...
 gemspec
 
-gem "rake", "~> 10.4"
-gem "rspec", "~> 3.3"
-gem "simplecov", "~> 0.11.0"
-gem "coveralls", "~> 0.7"
-gem "awesome_print", "~> 1.2"
-gem "pry", "~> 0.10"
-gem "factory_girl", "~> 4.5"
 gem "rubocop", "0.35.1", require: false
```
J3RN: I somehow missed this in my first pass -- is there a reason RuboCop wasn't moved to the gemspec?

tmuxinator.gemspec
J3RN: Seeing as factory-girl relies on activesupport anyhow, I suppose we'll leave that as-is. Thanks for looking into it and updating the CHANGELOG!

seanmalloy: I think I have everything updated now.

* Rebased my branch against master
* Updated CHANGELOG
* All gem deps moved to tmuxinator.gemspec (including rubocop)
* bumped required Ruby version to 2.0.0 in tmuxinator.gemspec
* tmux 2.2 and 2.3 added to TravisCI test matrix

J3RN:
Great work @seanmalloy! 👍

Issue #462, opened 2016-09-25
tmuxinator config files in a different location


emayssat-ms:
I would like to be able to change the location of the tmuxinator file
For example 
tmuxinator list --directory /tmp/tmuxinator 
should report the projects in /tmp/tmuxinator instead of ~/.tmuxinator

is something similar possible?

ethagnawl:
@emayssat-ms This is not currently possible. 

However, this feature could potentially be added in conjunction with #360.

ethagnawl:
@emayssat-ms This should now be possible on master via #511. This feature has not yet made it into a release of the gem, though.

Issue #360, opened 2015-11-29
XDG base directory support


Yamakaky:
See http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
For tmuxinator, it would mean use `~/.config/tmuxinator` instead of/in addition to `~/.tmuxinator/`. It should not be difficult to add.

suy:
To be pedantic: It should be `$XDG_CONFIG_HOME/tmuxinator`, with `$XDG_CONFIG_HOME` defaulting to ~/.config if unset. I think I can do this myself, if you are open to the change.

adamstrickland:
@suy i can only speak for myself, but I'm open to the change, but in all honesty I don't know much about XDG and all that.  Some questions that would need to be addressed may be:

- Besides conforming to standard, does this buy us anything?
- What cost (e.g. complexity, etc.) do we pay for this?
- How would this affect non-XDG operating systems?

In all honesty, while I am generally in favor of conforming to standards, I have bigger fish to fry, so I personally wasn't going to take this on (simple though it may be) in part because I don't know much about the context.  But I'm not against it either, should someone want to make a FOSS contribution.

geowa4:
For me, this would make my dotfiles repo much simpler. I'd essentially just have to clone into ~/.config.

ncoop:
Suggestion on how to implement:

1. Look at `$XDG_CONFIG_HOME` if it's set and not empty, or else `~/.config/`. If `tmuxinator` dir (no dot) is there, use it.
1. Else use `~/.tmuxinator` if it exists.
1. Else there's no config dir, so create create it in the first location.
1. If the first location doesn't exist, create it in home.

This is similar to [how git config searches](https://git-scm.com/docs/git-config#FILES), and it's nice because users who don't want to organize with XDG don't even have to know about it. (Git config also has a custom env var that has higher priority.)

With this done, I think `mux doctor` should also report where the config dir is.

J3RN:
@ncoop That sounds reasonable. In all honesty, I hate $HOME being crowded with dotfiles, so moving to using the `.config` directory seems ideal. I'll try to make time to start on this today.

Soliah:
There's one more case where tmuxinator also supports looking at the current directory for a `.tmuxinator.yml` file.

ncoop:
Today I learned there's a gem to facilitate this:
https://rubyworks.github.io/xdg/

HaleTom:
[...]

Issue #427, opened 2016-06-05
Introduce Analytics Module


ethagnawl:
Per [#426](https://github.com/tmuxinator/tmuxinator/pull/426#issuecomment-223695238), I'd like to start a conversation about introducing an opt-in analytics module.

Open questions:
1. how will data be communicated/persisted? (e.g. Net::HTTP => Firebase)
2. how will the opt-in feature fit into to the Gem installation workflow?
3. what information should be captured? (e.g. Tmux version, OS, OS version, misc. usage, etc.)

J3RN:
Here's my two cents:

1. I've never had fun with Net:HTTP. I'd lean towards [httparty](https://github.com/jnunemaker/httparty)  or [faraday](https://github.com/lostisland/faraday), but that's only my opinion.
2. Could we do this with a post-install trigger?
3. Tmux version and OS are a must. Naturally, the Tmuxinator version should be reported. I'm thinking that a request should be fired off whenever a user runs a certain command (mux start?) and should probably be run async so as to not cause a noticeable performance hit.

ethagnawl:
> I've never had fun with Net:HTTP. I'd lean towards httparty or faraday, but that's only my opinion.

Generally speaking, I agree. However, I'd prefer to avoid introducing a new dependency and Net::HTTP should be tolerable if we're only making a single or very few, similar requests.

> Could we do this with a post-install trigger?

Right - that's what I was thinking. However, after a (brief) look through the Gemspec docs, I don't see any mention of post-install hooks/triggers - just the infamous `post_install_message`.

The quick-and-dirty approach would be to have the gem prompt the user the first time it's run and then write their decision out to a file, which it'd reference in the future.

> I'm thinking that a request should be fired off whenever a user runs a certain command (mux start?) and should probably be run async so as to not cause a noticeable performance hit.

This is the one aspect of this feature I'm unsure about. It'd be nice to know which commands people are using, but I'm not entirely sure what's to be gained from that data.

I was thinking it might be good to start with a single request that's triggered post-install (which attempts to collect Tmux version, Tmuxinator version, OS, etc.) to see if people are even interested in sharing this data.

podung:
How about an opt-out instead?

I'm concerned about how many users would go out of their way to opt-in and question the value if we only get data from a small fraction of our user base.

An obvious, informative post-install message and frictionless opt-out seems like a reasonable approach (something like `mux optout`).  

[...]

Issue #440, opened 2016-07-21
Local config files use global path


ethagnawl:
The first line of the project config file generated when using the command, `tmuxinator new foo --local` contains a path to the global tmuxinator config directory. e.g. `~/.tmuxinator/foo.yml`, instead of `/some/directory/.tmuxinator.yml`.

I'm unaware of any problems this causes, but it's probably a bit confusing to the end user.

(It's also confusing that the local project requires a name that isn't ever used anywhere, but that's part of a larger conversation...)

PR #479, opened 2016-10-20
Support specified config directory and XDG Base Dirs Spec

HaleTom:
Allow setting config directory and honour XDG Base Dirs Spec

Resolve Issue #462 - "Tmuxinator config files in a different location"
Resolve Issue #360 - "XDG base directory support"
Mentioned in Issue #427 - "Introduce Analytics Module"

Allow `$TMUXINATOR_CONFIG` to specify a configuration directory, which
will be created (including its parent directories) if it doesn't exist.  This
directory takes preference over honouring the XDG specification (below). If
specified, this will be the _only_ configuration location used.

Honour the [XDG Base Directories Specification](https://specifications.freedesktop.org/basedir-spec/latest/)
with regards to user specific configuration files: Look for `tmuxinator`
configuration files under `$XDG_CONFIG_HOME/tmuxinator`.

For existing setups with no `$XDG_CONFIG_HOME` directory, continue to
create new projects in `~/.tmuxinator` for backwards compatibility.

If `$XDG_CONFIG_HOME/tmuxinator` exists, create new projects there.

If neither `~/.tmuxinator` nor `$XDG_CONFIG_HOME/tmuxinator` exist, use
`$XDG_CONFIG_HOME/tmuxinator` for new projects. Create this directory
(including its parents) if it does not exist.

When loading a project, search `$XDG_CONFIG_HOME/tmuxinator` before
`~/.tmuxinator`.  Document in comments and tests the existing behaviour
of returning only the first project file found in a recursive search of
a configuration directory.

Add `Config#directories`: an array of the configuration director{y,ies}
which exist. Update commands `implode` and `list` to operate upon this
array.

Add `alias`es for the 3 renamed methods below for backward
compatibility:

Rename `Config#root` -> `Config#directory`
This directory is the only one used when creating new project files, and
may be either `$TMUXINATOR_CONFIG`, `$XDG_CONFIG_HOME/tmuxinator` or
`~/.tmuxinator`.

Implementing XDG Base Dirs support with backwards compatibility means
that two directories may contain project files, making the `root`
nomenclature misleading as it implies a single directory.

Rename `Config#project_in_root` -> `Config#global_project`
See comment on `root` terminology above.  `global_project` is also
shorter and corresponds to the naming of `default_project`

Rename `Config#project_in_local` -> `Config#local_project`
Fit with the naming of both `global_project` and `default_project`

Sort gem dependencies into alphabetical order


HaleTom:
OK, ready for comments / merge.

adamstrickland:
This looks good 👍

adamstrickland:
@ethagnawl @J3RN do you have any feedback on this for @HaleTom?  I say LGTM.

ethagnawl:
@adamstrickland I've reviewed these changes a few times and, barring a few small nitpicks (adding comments now), everything _looks_ good to me. Though, I haven't had a chance to pull it down and experiment yet.

lib/tmuxinator/config.rb
```diff
@@ -4,18 +4,34 @@ class Config
     NO_LOCAL_FILE_MSG = "Project file at ./.tmuxinator.yml doesn't exist."
 
     class << self
-      def root
-        root_dir = File.expand_path("#{ENV['HOME']}/.tmuxinator")
-        Dir.mkdir(root_dir) unless File.directory?(root_dir)
-        root_dir
+      # The directory (created if needed) in which to store new projects
+      def directory
+        environment = ENV["TMUXINATOR_CONFIG"]
```
ethagnawl: This config get/test behavior is used in multiple places, so it might warrant its own helper method.

lib/tmuxinator/config.rb
```diff
@@ -75,9 +94,26 @@ def wemux_template
         asset_path "wemux_template.erb"
       end
 
+      # Sorted list of all .yml files, including duplicates
       def configs
-        Dir["#{Tmuxinator::Config.root}/**/*.yml"].sort.map do |path|
-          path.gsub("#{Tmuxinator::Config.root}/", "").gsub(".yml", "")
+        configs = []
+        directories.each do |directory|
```
ethagnawl: Is there a good reason not to use `map` here?

HaleTom: I guess a brain-fart doesn't count :)

lib/tmuxinator/config.rb
HaleTom: Thanks for the thorough review, I'll get on to these in a couple of days to allow for more reviewers to have their say in the meantime.

HaleTom:
@ethagnawl I trust that this is good to go now. Thanks for the thorough review - I learned from it.

J3RN:
This LGTM 👍

ethagnawl:
@HaleTom I had some time to experiment with this today. With the exception of one very small item (covered below), everything appears to work as intended.

I noticed that the config path at the top of the project yml file always uses the old default directory (~/.tmuxinator/$PROJECT). It's possible that this is caused by some legacy config on my system and it doesn't break anything (that I'm aware of, at least) but it's probably worth addressing. Also, as a bonus, the fix will likely address #440.

ethagnawl:
@Soliah @adamstrickland Any thoughts on the above? Shall we merge this as-is? It might be a bit self-serving, but I'd like to begin using XDG on my dev machines.

ethagnawl:
Merging this broke the build. I'm looking into it now and will submit a patch.

@HaleTom If you're available, would you mind taking a look to see if anything jumps out at you?

HaleTom:
I'm working towards diagnosing this. I got out of hospital yesterday (dengue), and am battling a `json` error when running `bundle install` under Manjaro, but I am trucking on...

HaleTom:
The doco for [`Dir::glob`](https://ruby-doc.org/core-2.2.3/Dir.html#method-c-glob) says:

> Note that case sensitivity depends on your system (so File::FNM_CASEFOLD is ignored), as does the order in which the results are returned.

I'll append `.sort` so that the results are (hopefully) system-independent.

HaleTom:
#506 created. I hope this is what you were after - let me know if not.

ethagnawl:
@HaleTom Wow. I hope you're feeling better! :|

Thanks for looking into this issue. Unfortunately, PR #506 doesn't look like it's going to address the problem _and_ reintroduce the changes from your original PR. Perhaps that's because #506 wants to merge its changes into the branch: tmuxinator:revert-479-Issues-360-462? Could you try creating a new PR and selecting `tmuxinator/tmuxinator base:master` as the target branch?

PR #480, opened 2016-10-21
Fix completions for ZSH

Xenopathic:
Some command descriptions were specified with Ruby word arrays, which
resulted in newlines being printed with `tmuxinator commands zsh`,
breaking completion. Also, `tmuxinator --help` showed weird whitespace
for these command descriptions. This commit runs a regex to replace
runs of whitespace with a single space for these command descriptions.


adamstrickland:
Hm...  we switched to using the string arrays in the descriptions pretty [recently](ab57f3734b7abd273a46ef6c00c2c7c96fda1852).  It sounds like those changes made behavior strange in`zsh`...  I'd rather not play whack-a-mole with this; @Xenopathic as a `zsh` user, would you be able to create a test or something that would allow us to verify results in that shell alongside `bash` and the others?

Xenopathic:
Ah, the issue is already fixed by ab57f3734b7abd273a46ef6c00c2c7c96fda1852, so these `gsub`s are unnecessary. I probably should have actually checked what version of tmuxinator was installed on my system before immediately making a patch based on git master...

PR #482, opened 2016-11-09
Add ability for pre_window commands to parse yaml arrays (as well as strings)

rpassis:
Currently, only `pre` commands accept yaml strings and arrays. This commit adds the ability for `pre_window` commands to also accept yaml strings and arrays.

### DIFF
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -135,7 +135,8 @@ module Tmuxinator
       elsif pre_tab?
         yaml["pre_tab"]
       else
-        yaml["pre_window"]
+        pre_window = yaml["pre_window"]
+        pre_window.is_a?(Array) ? pre_window.join("; ") : pre_window
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/tmuxinator/project.rb
```diff
@@ -135,7 +135,8 @@ def pre_window
       elsif pre_tab?
         yaml["pre_tab"]
       else
-        yaml["pre_window"]
+        pre_window = yaml["pre_window"]
+        pre_window.is_a?(Array) ? pre_window.join("; ") : pre_window
```
J3RN: Since this will be the third time this logic is used in this file, we should probably break it out into a private method and throw it at the bottom of the class somewhere.

lib/tmuxinator/project.rb
```diff
@@ -135,17 +131,14 @@ def pre_window
       elsif pre_tab?
```
J3RN: I just noticed that there are several conditional branches in this method. While multiple commands don't make sense for `rbenv` or `rvm`, would you mind also adding a call to `parsed_parameters` here with `yaml["pre_tab"]`?

J3RN:
@rpassis Thanks for making those changes! This LGTM 👍 I just wish I knew why Code Climate is so upset, but it won't tell me 🤔

Issue #476, opened 2016-10-17
tmuxinator and closing session


emayssat-ms:
I think it would be a great feature to be able to run commands when closing sessions, panes, or windows. That is in general adding hooks on tmux action/events. This is already what tmuxinator does on tmux start session, create windows, create pane. Currently when we kill a session I have no idea what happens to my running processes.

PR #486, opened 2016-12-04
Add tmuxinator project hooks

endorama:
Hello, this is my first contribution here. I found myself in need of more control around tmux session "events", and after looking at #476 I decided to provide this PR.

It adds 2 new "events" that can be acted upon in the project configuration file: on tmux session reattach and on tmux session stop.

A practical use case would be to run `vagrant suspend` and `vagrant up` on session detach and reattach.

Plus I duplicated the `pre` and `post` setting rewording it as `on_start` and `on_detach`, because so the code has been moved in the `Tmuxinator::Hooks` module and made sense with the new "hook" naming. As I don't know you policy about breaking changes, please let me know if and how you would like to address this.

Thank you for this little "gem" ;) it is really useful and has really helped me in managing my workflow.


adamstrickland:
@endorama ^ it looks like there's some extra whitespace in the spec.  Annoying, I know, but valuable for the maintainers. In the future you can run the rubocop checks locally with `bundle exec rake rubocop`

endorama:
I'm sorry, I did that but that one has slipped trough my checks. Everything should be fine now.

lib/tmuxinator/hooks.rb
```diff
@@ -0,0 +1,39 @@
+module Tmuxinator
+  module Hooks
+    def hook_on_reattach
```
ethagnawl: Perhaps a helper method could/should be made to reduce the duplication in this module? There could be a `supported_hooks` array which the helper could be mapped over. This would also make adding new hooks much simpler.

I realize that's potentially YAGNI, but I was just reading the tmux documentation and there are _many_ more hooks users might want to make use of:

> Each tmux command has a before hook and an after hook and there are a number of hooks not associated with commands.

endorama: I didn't propose such solution because I preferred to give something simpler to review as is my first contribution, but if that is the preferred path for this feature I'll be more than welcome in contributing on that!

Also be aware that this is a "hook on tmuxinator project events" and not a "hook on tmux event".  
AFAIU tmux hooks are global/session level configuration, while this PR simply expand the `pre` and `post` configurations already available.

I probably choose the wrong name "hooks" as is misleading in this context, and talking about *tmux session events* helped the misunderstanding. I'm sorry. I reworded the title to be more correct.

ethagnawl: @endorama Thanks for clearing that up. Let's revisit the proposed refactoring if/when additional _Tmuxinator_ hooks are added.

ethagnawl:
@endorama After our earlier exchange, I found a few minutes to pull this down and experiment.

A few thoughts:
- If others agree that the "hooks" approach is a better solution than sticking event methods in Project (I do), I'd propose aliasing the existing methods to their corresponding Hook methods, adding a deprecation warning and removing them in a future release.
- sample.yml is missing an `on_start` entry 
- Should the README's _Custom attachment and post commands_ section be updated to cover the new options and deprecations (if that's the approach we decide to take)?
- I think `hook_on_stop` _may_ need to be moved before `tmux_kill_session_command`. On my system (Debian 8.6, Bash 4.3.30, Tmux 1.9, Ruby 2.3.1p112) `on_stop` commands are ignored unless I reverse these two method calls. Is this working for you?
- Document (and potentially fix) behavior of on_reattach_hook; it appears to only be executed when reattaching via `tmuxinator start ...` and not `tmux attach -t ...`

endorama:
> sample.yml is missing an on_start entry

I'll fix this

> I think hook_on_stop may need to be moved before tmux_kill_session_command. On my system (Debian 8.6, Bash 4.3.30, Tmux 1.9, Ruby 2.3.1p112) on_stop commands are ignored unless I reverse these two method calls. Is this working for you?

It is. I'm currently using my fork to start and stop a vagrant machine when issuing `tmuxinator start project` and `tmuxinator stop project`

> Document (and potentially fix) behavior of on_reattach_hook. From what I can tell, it appears to only be executed when reattaching via tmuxinator start ... and not tmux attach -t ....

I would say this is by design. AFAIU for executing on `tmux attach` the proper tmux hook should be used. What is instead happening is that the `hook_on_reattach` command is written in the template file if [tmux has a session](https://github.com/tmuxinator/tmuxinator/pull/486/files#diff-c8dc30badcecf2dc19abb8a626398faeR80) named as the project name.

I'm starting to think that the naming *hook* is really misleading, in particular for `on_attach` and `on_reattach` ( which are more tmux concepts than tmuxinator ones ). Should I rename them in a more project-related way? For example:
- `on_reattach` could be `on_project_restart`
- `on_attach` could be `on_project_first_start`
- `on_start` could be `on_before_project_start`
- `on_stop` could be `on_project_stop`

This renaming would also be future proof when proper tmux hooks are implemented. Sorry for all this confusion.

ethagnawl:
> Should I rename them in a more project-related way? 

 :+1:

endorama:
Sorry for the delay.

- hooks remaned to `hook_on_project_*`
- extracted the common code for hooks
- moved project hooks to `Tmuxinator::Hooks::Project` to make the scope clearer
- updated relevant specs

ethagnawl:
@endorama Sorry for the delay. I'll have to find some time to revisit this and do a proper review. Did you give any additional consideration to adding a deprecation warning to the existing pre/post "hooks"?

In the meantime, does anyone else have any thoughts or feedback? Is this something people think is worth pursuing? If so, are these hooks deserving of a README entry?

/cc @J3RN @adamstrickland @Soliah

Soliah:
I'll review this on the weekend.

endorama:
@Soliah thanks!

@ethagnawl IMHO, `pre` and `post` could be deprecated, as I think these "hooks" are more explicit and perform the same action, but I'd leave this decision to you as you see fit.

Soliah:
Ok PR looks, good. Definitely think the `pre` and `post` hooks should be deprecated. Additionally maybe lets cut a pre release version for testing in the wild.

endorama:
Hello and sorry for the delay! I deprecated the `pre` and `post` options.

endorama:
Hello @Soliah, any news about this? Just to confirm I'm being using this patch since February.

endorama:
:tada: :tada: 

Thanks a lot!!! :smile:

TTycho:
>  Hello and sorry for the delay! I deprecated the pre and post options.

The `pre` option could be set per-window. As I understand from the above, the hooks or `pre_window` option do not function like that. Is there an alternative for `pre` in this case?

PR #497, opened 2017-01-10
Updated details on supported tmux versions

J3RN:
The range of versions I listed reflect what we test against on Travis.

Thoughts @adamstrickland @ethagnawl?


adamstrickland:
Agree

Soliah:
Hrm, I wouldn't completely trust CI in this case on versions prior to 1.8 as we're not really running any integration tests against tmux itself. We're really just asserting against the generated shell script but I think we'd want some proper integration tests against tmux to confidently support versions prior to 1.8.

A good integration test would be to actually run tmuxinator and assert against the state of the session.

Soliah:
I have some ideas on how we could test against tmux. 

- Use standard `list-sessions`, `list-windows` and `list-panes` to test for session structure
- Use `capture-pane` to write the buffer to a file and assert against command outputs

Will have to try a PoC for this. There is also an allow_failures for 1.7 in the travis confit which will need to be removed.

adamstrickland:
Holding off on this until we have a better integration test scheme

PR #501, opened 2017-01-27
Fix typos

rglassett:
Noticed a typo in `sample.yml` when I was installing. Went through the code with `aspell` to clean things up.


J3RN:
Hey @rglassett! Thanks for your PR! :sparkles:

Whoa! Thanks for catching all of those! I think this PR looks great as-is. 👍 

Thoughts @Soliah @adamstrickland?

PR #502, opened 2017-01-29
First pass at some basic integration tests

Soliah:
Added first integration test that actually executes the project and launches a tmux session which gets asserted against.

Approach is what I suggested in #497. We would still need more coverage than this simple project though.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,7 +24,7 @@ RSpec.configure do |config|
   config.order = "random"
 
   # Don't run integration tests by default
-  config.filter_run_excluding integration: true
+  # config.filter_run_excluding integration: true
 end
 
 # Copied from minitest.
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -21,6 +21,9 @@ require_relative "matchers/pane_matcher"
 
 RSpec.configure do |config|
   config.order = "random"
+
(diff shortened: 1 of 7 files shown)


adamstrickland:
Updated branch.  I agree that this falls short of what we *want* but maybe we merge it in to at least get the ball rolling?

spec/spec_helper.rb
```diff
@@ -22,6 +22,9 @@
 
 RSpec.configure do |config|
   config.order = "random"
+
+  # Don't run integration tests by default
+  # config.filter_run_excluding integration: true
```
ethagnawl: Should this line be commented out?

ethagnawl:
Looks like a _good enough_ start to me!

Soliah:
Hopefully the direction is clear - the implementation isn’t particularly good and I haven’t had time to flesh it out more.

spec/spec_helper.rb
Soliah: I think these tests should be run separately maybe so yes.

Soliah:
Also from memory the most difficult part was determining the state of the tmux session to assert against - especially across versions. capture_pane and stuff are only available in certain versions.

adamstrickland:
Looks like this still needs some work.  Prolly should hold off for now.

ethagnawl:
Should we close this for now?

PR #507, opened 2017-02-07
Attempt at salvaging PR #479

ethagnawl:
Some background:

After #479 was merged into master, the CI job began failing. I have to admit that I don't understand why this would have happened. It's possible that the test results on the branch were out of date and after I fixed what I thought were two trivial merge conflicts, the tests were re-run, the failures started and I overlooked them. However, those same changes locally did not result in any test failures. 

Either way, after that went down, I reverted the botched PR using GitHub's "Revert" feature. Then when @HaleTom tried to re-open the PR by making a patch which attempted to fix the aforementioned failures on their branch, the original changes were being ignored because of the "Revert" commit.

This is an attempt to force Git to recognize the original changeset. The only solution I was able to come up with was checking out the original branch, applying the patch and then squashing the history. (If folks care about the branch history and know of a solution which preserves the commit history on the branch, I'd be all for it.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ rvm:
   - "2.2.3"
 env:
   - TMUX_VERSION=master
+  - TMUX_VERSION=2.3
+  - TMUX_VERSION=2.2
   - TMUX_VERSION=2.1
   - TMUX_VERSION=2.0
   - TMUX_VERSION=1.9
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Unreleased
 ### Misc
 - Removed support for Ruby 1.9.3
+- Move gem dependencies from Gemfile to tmuxinator.gemspec
+- Add tmux 2.2 and 2.3 the TravisCI test matrix
 
 ## 0.9.0
 ### Misc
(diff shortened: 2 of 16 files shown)


HaleTom:
I'm glad it wasn't just me having problems with the previously merged changed being hidden. Nice solution. 

*following*

adamstrickland:
So I'm going to vote that we ignore the CodeClimate issue so we can LGTM.  Thoughts?

ethagnawl:
@adamstrickland Would you like to do the honors? I'm feeling a bit skittish after my last attempt.

HaleTom:
Won't this error just keep reappearing on every new build, or can it be disabled permanently?

How do I get the complexity rating on my own machine? I can try to "decomplexify" it, and remove the unidiomatic `return`.

ethagnawl:
@HaleTom Good point/questions. I imagine the issue will persist...

You could either pull my branch, make the change and submit a new PR (kicking off a new Code Climate run) or download the [Code Climate CLI](https://github.com/codeclimate/codeclimate) and try running that against the project locally*. I don't see a Code Climate config file anywhere, so I'm guessing it's using the default configuration. If you prefer the latter, let me know and I can update my PR if your fix checks out. 

FYI, I tried simply removing the `return` and that resulted in a test failure.

```
expected: ("TMUXINATOR_CONFIG")
got: ("XDG_CONFIG_HOME")
Please stub a default value first if message might be received with other args as well. 
```

\* I've never tried this, so maybe someone else can chime in regards to its feasibility. /cc @Soliah @adamstrickland @J3RN

Issue #520, opened 2017-04-24
"Failed to parse config" provides no useful details


crobicha:
Is it possible to print either the line number or the yaml error message when there is a problem with the config? Seems like this would make debugging a lot easier. Earlier I wasted several hours debugging a failed config only to realize I had added a line with tab instead of spaces.

I am running tmuxinator 0.9.0 with tmux 2.3 on both OSX and Debian Linux. I could look into this if someone could point me in the right direction, thanks!

ethagnawl:
This is a great idea, @crobicha! 

I just added an exception variable and `puts` to the relevant `begin/rescue` block and the error message generated by my test scenario (replacing `-` with `x`) was intelligible and pointed directly to the line/column containing the error. (If you'd like to share your broken config, I'd be interesting in seeing the error message the parser generates for it - I tried to mix up spaces/tabs in my project but the Yaml was still parsed without issue 😕.)

If you're up for it, you're welcome to take a pass at implementing this. The `rescue` in question lives [here](https://github.com/tmuxinator/tmuxinator/blob/master/lib/tmuxinator/project.rb#L44-L46). Just review [CONTRIBUTING.md](https://github.com/tmuxinator/tmuxinator/blob/master/CONTRIBUTING.md) and have a look through some of the other classes/tests to see how this sort of logging is handled and tested (or not).

crobicha:
Thanks, I've got a simple fix working locally, just changing that line to

> raise "Failed to parse config file: " + e.message

What would be a simple way to test this? Maybe an invalid yml file and check for error message when trying to start a new session?

ethagnawl:
> What would be a simple way to test this? Maybe an invalid yml file and check for error message when trying to start a new session?

That sounds reasonable to me.

/cc @Soliah @J3RN @adamstrickland

crobicha:
I can't seem to find the proper way to get the rake test working on my setup, so I published the branch with the functional change (issue/520-config-parse-error-message). Maybe someone could add a quick test but for me I can verify it is working locally

ethagnawl:
@crobicha I will take a look.

ethagnawl:
@crobicha Which version or Ruby are you using?

ethagnawl:
@crobicha Please see #521.

crobicha:
Thanks for taking care of this! The existing project tests were passing, I just was having issues writing the new test case, probably due to operator error :-)

PR #521, opened 2017-05-07
Issue/520 config parse error message

ethagnawl:
Addresses #520.


ethagnawl:
@adamstrickland Would you like to do the honors?

Issue #524, opened 2017-05-19
Suggest updating supported ruby versions


adamstrickland:
Support for Ruby 2.0 and 2.1 has ended.  Suggest moving supported versions to 2.2, 2.3 and 2.4.  Further, suggest a policy to only support ruby versions that are themselves currently supported as a project policy.  Thoughts?

adamstrickland:
@tmuxinator/tmuxinator any thoughts on this?  While we're on the topic, do we feel there's a reason to (or not to) deprecate support for some of the older versions of tmux, as well? We currently support back to tmux 1.5, which was released 6+ years ago.

ethagnawl:
Keeping pace with the officially supported versions of Ruby sounds reasonable. However, I'm not aware of any issues which have surfaced as a result of people using legacy versions. It seems like the major downsides of continuing support for legacy versions are: the library can't make use of some of the newer features (e.g. squiggly heredocs in `Cli::COMMANDS`) and the additional time added to the CI job.

I feel more strongly about dropping support for legacy versions of tmux because there are (at least two) conditionals which test for version ranges. 

I don't have much to offer in terms of suggesting which versions of tmux we drop support for - this calls to mind #427. Though, one data point which might be worth mentioning is that the latest version of Debian (Stretch) offers 2.3 and the most recent "obsolete stable release" (Jessie) offers 1.9-6. Depending on what other distributions/OSs are offering, drawing the line at 1.9* would seem reasonable to me.

ethagnawl:
Per #538 and @adamstrickland's [comment](https://github.com/tmuxinator/tmuxinator/issues/536#issuecomment-307885272) in #536, perhaps we should also consider adopting an upper bound on the supported version of tmux?

techgaun:
I feel at least, the `master` branch deserves a basic support in terms of supporting existing commands. My problem specifically was that due to the version check, the tmux master was generating command that're supposed to be the commands for older version of tmux. Its of course not always possible to support new additions or changes but it would at least be great to default to the newer command sets.

alexkrolick:
Mac OS ships Ruby 2.0 as the system Ruby, so this is kind of annoying. Is there a technical reason to require a higher version?

ethagnawl:
We've been tracking the supported Rubies for a while now and have updated the docs to reflect that. Feel free to reopen if there's anything else to discuss.

PR #526, opened 2017-05-20
Issue 524 update ruby versions

adamstrickland:
Suggested changes for #524


adamstrickland:
Does anyone mind if I merge this in to master?  This (i.e. updating our supported rubies) should be done regularly.  The question is, do we want to (continue to) support versions of ruby for tmuxinator that are no longer themselves supported?

J3RN:
@adamstrickland FWIW: Yes, I think we should merge this :+1: and should do so regularly. No, I don't think we should support rubies that the ruby maintainers have decided to no longer support. I hope, for security reasons, mostly, that no one is using an unsupported ruby. Besides, one can install a supported ruby and tmuxinator with it while using an unsupported ruby for work, etc.

ethagnawl:
@adamstrickland Supporting only modern Rubies sounds reasonable to me. 

Out of curiosity, how do we determine which patch version to support/test against? Should it always be the most recent stable release? FWIW, the three versions specified in .travis.yml and tmuxinator.gemspec are all one patch behind the most recent stable versions.

adamstrickland:
@ethagnawl I chose the 3 most-recent stable versions at the time I submitted the PR.  As you noted, that has since changed; I have bumped the versions accordingly.

In general, I think we should simply try to stay current, bumping to the most recent stable on a regular basis, but IMO we don't need to watch it like a hawk, either.

PR #538, opened 2017-06-27
fix default path for tmux master

techgaun:
I'm using tmux master branch with tmuxinator and get the error reading `invalid option default-path`. The version has `to_f` so for something like `tmux -V` that returns `tmux master`, we get 0.0 and thus the `default-path` is generated instead of `-c`. I feel this is not very pretty but I hope this is acceptable in terms of real usage.

### DIFF
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -53,7 +53,7 @@ module Tmuxinator
       end
 
       def default_path_option
-        version && version < 1.8 ? "default-path" : "-c"
+        version && version < 1.8 && version > 0.0 ? "default-path" : "-c"
       end
 
       def editor?


ethagnawl:
@techgaun Thanks for submitting this PR. Though, I'd ask you to please see the conversation in #524 regarding officially supported versions of tmux. Feel free to chime in, as you can likely provide valuable feedback.

techgaun:
@ethagnawl thanks for the response.. I did see that issue and didn't see anything about the tmux master branch. It seems that the tmux master branch always has `tmux master` as the response of `tmux -V` (https://github.com/tmux/tmux/blob/411640c0327525f2e5e6d61bd81623ce64d0d811/configure.ac#L3). It would be really good to have tmux master branch handled at least from the perspective of generating the basic commands.. this way, I think it opens ease for adding new changes earlier on tmuxinator side as well. I will read the conversation on that issue again more carefully. thanks!

raiyuza:
Disagreed, released versions should have version dependcies not bases on ~HEAD or branch_name. 

This issue should be closed.

techgaun:
sounds good. closing

Issue #457, opened 2016-09-21
Move validation functionality out of Tmuxinator::Config


ethagnawl:
As discussed in #452, we should consider moving any validation functionality (`validate`, `installed?`, `editor?`, `shell?`, `exists?`, etc.) out of `Tmuxinator::Config` and into a new `Tmuxinator::Validator` module.

J3RN:
Since these methods are all related (exclusively?) to the "doctor" command, maybe the module/class should be called `Tmuxinator::Doctor`.

PR #541, opened 2017-07-13
Extract Tmuxinator::Doctor utility class

rglassett:
Extracts methods from Tmuxinator::Config that were only invoked as part of the `tmuxinator doctor` command.

https://github.com/tmuxinator/tmuxinator/issues/457


adamstrickland:
:+1:

PR #542, opened 2017-07-14
[Fixes #485] Respect global pane-base-index option

### LINKED ISSUE #485: Can't find pane 0
After starting a session I get a `can't find pane 0` [message](https://youtu.be/tnv0LzaI4Os) in the console. I don't observe any other issues then the message: windows and panes are created, and programs are started. I think the bug is related to the fact I change `base-index` and `pane-base-index` to `1`, since when I remove the settings, the message doesn't appear.

Linux (Arch); tmux 2.3; tmuxinator 0.9.0; [configuration files](https://gist.github.com/Shamaoke/db76b4419c1d43b897d588a73caa19ac)

rglassett:
Tmuxinator currently fails to find a value for this option, because pane-base-index is a window option, not a session option: the -w flag must be passed to 'tmux show-option' in order to display it. This patch fixes the issue with two changes to the #show_tmux_options method of Tmuxinator::Project:

1. It now displays both global session options and global window options.
2. It now **only** detects the options that Tmuxinator::Project knows and cares about (currently base-index and pane-base-index)


adamstrickland:
Is there a spec that could be added to insure that this fixes the issue?

rglassett:
A red-to-green test would require changing how `tmux_config` is mocked. Right now the private method `show_tmux_options` is stubbed out and left untested; the tests don't even cover the method I've changed. I'd *like* to mock the `tmux` CLI itself and stub its responses to the `show-option` and `show-window-option` commands, but I'm not sure of the best way to do that.

adamstrickland:
In other places I have mocked the calls to `tmux` using this:

```ruby
allow_any_instance_of(Kernel).to receive(:`).with(command).and_return(output)
```

See https://github.com/adamstrickland/tmuxinator/blob/24b0330aa325d0c7efc3bccdd4fde1e5d097cb03/spec/lib/tmuxinator/config_spec.rb#L149-L153

rglassett:
Nice technique, thanks!

rglassett:
Any other action needed from my end to move this forward?

PR #543, opened 2017-07-14
Decouple base_index from pane_base_index

rglassett:
The pane_base_index and base_index options are set and handled independently in tmux, so the value of pane_base_index should not override the value of base_index (as it previously was).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Refactor Tmuxinator::Config by extracting a Tmuxinator::Doctor class (#457)
 - Fix a bug where startup_window and startup_pane were not respected if running
   tmuxinator from within an existing tmux session (#537)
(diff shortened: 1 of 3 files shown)


adamstrickland:
Is this a fix for a particular issue (I assume so...)? If so, I would expect a spec that models that specific problem; maybe the `project_spec` does so sufficiently, but it's not quite clear to me...

rglassett:
On current `master` there's no issue, since `pane-base-index` isn't properly detected/set in `tmuxinator`. But if/when #542 is merged, the user's `pane-base-index` option, if set, will always clobber their `base-index` option. I'll look at the specs a little more closely and see if there's a better test case I can add.

rglassett:
I've added a test case that fails in `master` and passes here.

rglassett:
Anything else needed before merging?

PR #544, opened 2017-07-14
[Fixes #537] Use full paths for startup window/pane

### LINKED ISSUE #537: Wrong window selected on startup, startup_window not working
By default, the startup_window should be the first one.

When starting the session, it is always the last window that is selected (in this case, server).

startup_window does not have any effect.

I am using:
tmux 2.5-1
tmuxinator 0.9.0-1
archlinux

Below is my configuration, in case it may be useful.
```
# ~/.tmuxinator/blog.yml

name: blog
root: ~/work/projects/nicolaswebdev_blog

# Specifies (by name or index) which window will be selected on project startup. If not set, the first window is used.
startup_window: 1

windows:
  - editor: vim
  - extra1:
  - extra2:
  - server: bundle exec jekyll liveserve --drafts
```

rglassett:
The startup_window and startup_pane options were not behaving correctly for users running tmuxinator from within an existing tmux session.  Tmuxinator would generate commands like `tmux select-window -t 1`, which would target the current session instead of the target (project) session. This bug is fixed by specifying full paths (`session:window.pane`) for the startup_window and startup_pane.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 - Remove unused attr_readers from Tmuxinator::Window
 - Add ability for pre_window commands to parse yaml arrays
 - Refactor Tmuxinator::Config by extracting a Tmuxinator::Doctor class (#457)
+- Fix a bug where startup_window and startup_pane were not respected if running
+  tmuxinator from within an existing tmux session (#537)
 
 ### Misc
 - Removed support for Ruby 1.9.3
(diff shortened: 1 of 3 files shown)


rglassett:
@adamstrickland how do you feel about merging this now that it's confirmed to resolve #537?

adamstrickland:
@rglassett thx for the reminder. Merged

PR #545, opened 2017-07-14
adds specific version for when tmux's version is 'master' (99.9)

adamstrickland:
In https://github.com/tmuxinator/tmuxinator/issues/524#issuecomment-312065082, commenter @techgaun makes a good point about `tmux` `master` not failing (IIRC there was an issue for this item specifically, but I can't seem to find it ATM).  This creates an admittedly-arbitrary numeric value (99.9) to enable downstream invocations of `Tmuxinator::Config.version` to return a sensible value 

## Context

Currently, if `tmux` is built from scratch from the `master` branch, `tmux -V` returns `tmux master`, which resolves numerically to `0.0`.  This is both factually incorrect (as of this writing, `tmux`'s master branch is based off of `2.5`) and misleading (as it implies that `master` is an older version than, say `2.5`).  I'm open to other suggestions on how to deal with this, but figured this is a relatively easy solution (again, I think I saw someone else suggest it; they deserve the credit).

The use of 99.9 is, as indicated, clearly arbitrary.  My thinking is that it's fair to assume that `master` should be considered as equal-to-or-greater-than the highest published version, and that, for `tmuxinator`'s purposes, as detecting same is to handle "API" changes as `tmux` ages, this should be sufficient.  Again, open to other ideas here...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Remove unused attr_readers from Tmuxinator::Window
 - Add ability for pre_window commands to parse yaml arrays
 - Refactor Tmuxinator::Config by extracting a Tmuxinator::Doctor class (#457)
+- Refactors Tmuxinator::Confir.version to display a high value number for `tmux`'s `master` branch
 
 ### Misc
 - Removed support for Ruby 1.9.3
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -3,6 +3,7 @@ module Tmuxinator
     LOCAL_DEFAULT = "./.tmuxinator.yml".freeze
     NO_LOCAL_FILE_MSG =
       "Project file at ./.tmuxinator.yml doesn't exist.".freeze
+    TMUX_MASTER_VERSION = 99.9
 
     class << self
       # The directory (created if needed) in which to store new projects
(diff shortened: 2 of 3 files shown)


rglassett:
I don't *actually* expect `tmux` to publish a version higher than 99, but `Float::INFINITY` might be a slightly better arbitrary value.

Is it correct to treat `master` as a numeric version at all, though? Depending on how frequently the user pulls from upstream, `master` may be ahead of or behind any given release.

adamstrickland:
Good point re: `Float::INFINITY`.  Will hold off on implementing until we decide the larger issue: is this a good plan, or should we do something else (as @rglassett states, perhaps it's not a great idea)?

ethagnawl:
> Depending on how frequently the user pulls from upstream, master may be ahead of or behind any given release.

I think this is an important point. 

Also, as I mentioned in #524, trying to support bleeding-edge master is problematic. Since Tmuxinator's test suite won't have been run against the latest feature set, there's no way to guarantee compatibility.

What would you all think about adding a supported version range check (again, per #524) and allowing people to use unsupported versions, but printing a warning message which tells them _here be dragons_ if they're outside of the range?

rglassett:
A version range check with warnings sounds reasonable to me, and it seems fine to treat `master` as an arbitrarily high version number in that context.

adamstrickland:
Closed in favor of https://github.com/tmuxinator/tmuxinator/pull/574

Issue #534, opened 2017-05-26
Add Configuration section to README


ethagnawl:
It's probably worth adding a Configuration section to README.md and outlining the various configuration options ($XDG_CONFIG_HOME and $TMUXINATOR_CONFIG) now/soon to be available.

ethagnawl:
Closed via #549.

PR #549, opened 2017-08-17
534 - README Entry for Project Configuration Location

ethagnawl:
Add _Project Configuration Location_ section to README.md.

Addresses #534.


ethagnawl:
> I haven't been following closely but does setting one of these env vars override the local .tmuxinator.yml behaviour?

@Soliah No, I don't believe so. I'll actually remove the reference to local project files.

PR #556, opened 2017-09-21
Use correct tmux socket in tmux_has_session

jajm:


augustelalande:
@akofink I think this can be closed since we now incorporate the socket into the tmux_has_session call
https://github.com/tmuxinator/tmuxinator/blob/25c2134f7dad04a7630f98c5b1e97981ff0dcd20/lib/tmuxinator/project.rb#L158-L180

akofink:
Agreed! Thanks for pointing this out 👍

PR #559, opened 2017-09-21
Remove confusing README section on base indexes

rglassett:
Remove confusing README section about the pane-base-index and window-base-index options. These options can be set independently of one another now that #542 and #543 are merged.


ethagnawl:
@tmuxinator/tmuxinator Shall we merge this?

Issue #566, opened 2017-10-19
invalid option: default-path


JamesYuan:
Hi,
Only one windows is shown, and reproduce it like this.

`
$ # just new a demo, save and quit

$ tmuxinator new demo

$ # start it, only 1 window is shown, "kill-server" in the tmux

$ tmuxinator demo
invalid option: default-path
can't find pane 2
can't find window 2
can't find window 3
[server exited]
`
`
$ tmuxinator -v
tmuxinator 0.10.0

$ brew info tmux
tmux: stable 2.6 (bottled), HEAD
Terminal multiplexer
https://tmux.github.io/
/usr/local/Cellar/tmux/HEAD-45ee118 (10 files, 689.4KB) *
  Built from source on 2017-08-02 at 11:43:21
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/tmux.rb
==> Dependencies
Build: pkg-config ✔
Required: libevent ✔
Optional: utf8proc ✔
==> Options
--with-utf8proc
	Build with utf8proc support
--HEAD
	Install HEAD version
==> Caveats
Example configuration has been installed to:
  /usr/local/opt/tmux/share/tmux

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
`

Thanks,
James.

JamesYuan:
I just uninstall and install the tmux again, it works well now.

coddingtonbear:
I *think* what your re-installation might've done is instead downgraded you to a version of tmux before this change: https://github.com/tmux/tmux/blob/37f83adca8b965133ebdaeebea790ae897bd7175/CHANGES#L517

PR #567, opened 2017-11-03
When tmux is installed from a github clone, assume the later behavior rather than the earlier.

coddingtonbear:
Were a user to install Tmux from a clone from github, the version returned is not a version number:

```bash
$ tmux -V
tmux master
```

When attempting to convert the string `master` into a valid version number, Ruby's `to_f` returns `0.0`; this causes Tmuxinator to assume that the version number is _earlier_ than 1.8 (released in 2014) rather than later.  It seems reasonable to assume that were `0.0` returned, that the user is using a recent clone.  

This is the root of at least one existing issue: https://github.com/tmuxinator/tmuxinator/issues/566

### DIFF
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -49,7 +49,7 @@ module Tmuxinator
       end
 
       def default_path_option
-        version && version < 1.8 ? "default-path" : "-c"
+        version > 0 && version < 1.8 ? "default-path" : "-c"
       end
 
       def exists?(name)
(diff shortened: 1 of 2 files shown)


rglassett:
Attempts have been made to solve this problem in #538 and #545. I think I like the approach in #545 best. I might try to pick up @adamstrickland's existing work and round it out with some of the ideas discussed in that thread.

coddingtonbear:
You can totally feel free to close this PR, @rglassett; I submitted this just to quickly solve the problem in a way others might be able to use without knowing that there were existing solutions -- if you'd rather go with a different solution, go for it!

adamstrickland:
Closed in favor of https://github.com/tmuxinator/tmuxinator/pull/574

PR #573, opened 2017-11-22
Fix additional pane-base-index bugs

rglassett:
Fixes remaining sites where the base-index option (for windows) was incorrectly used in place of the pane-base-index option. This was causing some commands to be mis-targeted as I was setting up new projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 - Handle emojis in project names (#564)
+- Fix remaining sites where the base-index option (for windows) was incorrectly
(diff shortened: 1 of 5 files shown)


rglassett:
I agree that we should release this soon. What is your typical process? Commit to update CHANGELOG.md and then release the new tag (in this case, v0.10.1)?

adamstrickland:
@tmuxinator/tmuxinator does anyone disagree with cutting a patch release from master to repair this bug?  I'll wait a couple of days for comment; if there are no objections, I'm happy to do a release on Tuesday

PR #574, opened 2017-11-22
Handle 'tmux master' version more gracefully

rglassett:
Same diff as #545 with minor changes as discussed in that thread. Treats 'tmux master' as being arbitrarily high, and displays a deprecation warning to users attempting to run tmuxinator under 'tmux master'.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ rvm:
   - "2.3.5"
   - "2.4.2"
 env:
-  - TMUX_VERSION=master
   - TMUX_VERSION=2.3
   - TMUX_VERSION=2.2
   - TMUX_VERSION=2.1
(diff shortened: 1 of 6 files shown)


rglassett:
It occurs to me that I should link related issues in the commit message and changelog. I'll add that when I can.

rglassett:
It also occurs to me that the warning should be printed for **any** unsupported version, not just `master`. I'll go back and clean that up as well.

rglassett:
Comments addressed. Ideally I'd like `.travis.yml` and the Tmuxinator code to share a single source of truth for supported tmux versions, but it's probably not critical for now.

aepyornis:
@rglassett @adamstrickland @ethagnawl thanks for this! It solved the very problem I was having.

I was wondering how you'd feel about having an option - environment variable? cli flag? -- to disable the warning? That way those using tmux master wouldn't have to hit 'enter' every time they fire up tmuxinator.

rglassett:
:+1:, I think a `--quiet` flag makes sense.

PR #578, opened 2017-11-28
updates rubocop to resolve security vulnerability

adamstrickland:
per https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-8418

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,7 +69,7 @@ Style/CaseIndentation:
   Description: Indentation of when in a case/when/[else/]end.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#indent-when-to-case
   Enabled: true
-  IndentWhenRelativeTo: case
+  EnforcedStyle: case
   SupportedStyles:
   - case
   - end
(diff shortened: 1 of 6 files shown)


.rubocop.yml
```diff
@@ -876,6 +872,8 @@ Style/UnneededPercentQ:
   Description: Checks for %q/%Q when single quotes or double quotes would do.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#percent-q
   Enabled: true
+  Exclude:
+    - tmuxinator.gemspec
```
rglassett: Is the gemspec file "special" enough that we should just exclude it under `AllCops`, instead of a handful of individual cops?

lib/tmuxinator/project.rb
```diff
@@ -184,7 +184,7 @@ def tmux_has_session?(name)
       # Please see issue #564.
       unescaped_name = name.shellsplit.join("")
 
-      !!(sessions =~ /^#{unescaped_name}:/)
+      !(sessions !~ /^#{unescaped_name}:/)
```
rglassett: Weird that Rubocop now prefers for this to be phrased as a double negative, I'd thought `!!` was a well-established Ruby idiom. I'd probably rather drop the `!`s entirely and use the return value of `=~` (first matching index or nil). Not a huge deal, though.

Issue #582, opened 2017-12-05
Added a title to a pane


chrisspen:
I'm trying to run multiple panes containing `htop` being run through `ssh` to different servers. My configuration looks like:

    project_name: Server Monitor
    windows:
      - editor:
          layout: main-vertical
          panes:
            - ssh -t -i mykey.pem user@server1 htop
            - ssh -t -i mykey.pem user@server2 htop
            - ssh -t -i mykey.pem user@server3 htop

And when I launch it from tmuxinator, it runs the commands just fine. However, it obscures the SSH title, making it nearly impossible to tell which pane corresponds to which server.

Is there any way to tell tmuxinator to add some sort of "label" or "title" to each pane so I know which is which? I've read through the docs, but I don't see this feature mentioned anywhere.

chrisspen:
This does seem to be supported in tmux, as shown [in this example](https://stackoverflow.com/a/37602055/247542), but it's unclear how to access this feature through tmuxinator.

raiyuza:
Do we want this feature, cause you are able to do it with some ENV variable magic and scripts checking in background and adjust the pane names accordingly!

Abdull:
Having such feature in tmuxinator without some ENV magic is exactly the reason why I am using tmuxinator in the first place.

hiredgunhouse:
:+1:

ethagnawl:
#586 would add support for this, but last I checked it wasn't working as expected. I'll try to revisit the PR this week, but it'd be great if you all could take a look and provide any feedback you may have.

yorkbai:
Add renamePane() func in ~/.zshrenv first, then invoking it in yml file , can modify pane name but  inconvenient
```
# cat ~/.zshenv
function renamePane () {
  paneName=$1
  printf "\033]2;%s\033\\r:r" "${paneName}";

# less sample.yml
windows:
  - win1:
      layout: main-vertical
      panes:
        - renamePane()  “name” ;  ssh user@server1
```

ethagnawl:
Until this lands via #586 the desired behavior can be accomplished using the following:

```
# tmux config file

# the rest of your config file ...

# This example creates a variable named `custom_pane_title` which will be 
# displayed if and only if it has been set -- in this case by by your 
# tmuxinator pane commands.
# This feature doesn't seem to be documented, otherwise I'd provide a link. 
# I _believe_ this feature requires tmux >= 3.0
[...]

PR #586, opened 2017-12-16
Add Support for Pane Title

mehar:
windows:
  - w1:
       panes:
         - pane_title:
            - ls
         - pwd
         - git status

If the pane yml is a Hash, then we can set the pane_title variable, which in turn can be used in "pane_border_format"

Ex: setw -g pane-border-format " [#{pane_index}:#{pane_title}#{?pane_synchronized, SYNC,}] #{pane_current_command} "

### DIFF
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -59,6 +59,9 @@ cd <%= root || "." %>
         <% if pane.tab.pre %>
   <%= pane.tmux_pre_command %>
         <% end %>
+        
+        <% if pane.title %><%= tmux  %> select-pane -t <%= pane.tmux_window_and_pane_target %> -T <%= pane.title %><% end %>
+
         <% pane.commands.each do |command| %>
   <%= pane.tmux_main_command(command) %>
         <% end %>
--- a/lib/tmuxinator/pane.rb
+++ b/lib/tmuxinator/pane.rb
@@ -1,12 +1,13 @@
 module Tmuxinator
   class Pane
-    attr_reader :commands, :project, :index, :tab
+    attr_reader :commands, :project, :index, :tab, :title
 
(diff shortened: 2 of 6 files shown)


mehar:
screen shot 2017-12-17 at 9 50 57 am

bedge:
Very cool, would also like this. 
Allows one to fix all the panes named with unhelpful (from an ID standpoint) wrapper commands like ssh.

mehar:
Thanks @bedge

sabrehagen:
@ethagnawl This would be a really useful feature. Can this be merged please?

ethagnawl:
Hi, @mehar. Apologies for the delay on this.

I've pulled your changes and done some poking around. Does this actually set the pane title? When trying to do so, I see: 

```
tmux: unknown option -- T
usage: select-pane [-DdegLlMmRU] [-P style] [-t target-pane]
```

mehar:
@ethagnawl let me check

ethagnawl:
That codeclimate failure is annoying. Does anyone have any ideas for a simple solution? (Ideally, something other than just reverting [my changes](https://github.com/tmuxinator/tmuxinator/pull/586/commits/7da7e6f835d78715091ff9ddfc7f2fe2b4687611).)

ethagnawl:
@mehar Were you ever able to verify that this is working as expected?

woodgear:
is there some update here?

ethagnawl:
@mehar Any updates on this? Also, is the pane title set by the config option likely to be overwritten by tmux when the user does something which causes tmux to change the `pane_title` value?

ethagnawl:
While this approach does work initially, the pane title is not persistent. 

I think a better way forward would be to use a [user provided variable and `pane-border-format`](https://github.com/tmuxinator/tmuxinator/issues/582#issuecomment-625587104).

PR #587, opened 2017-12-27
Allow YAML aliases to work

agios:
This functionality needs to be explicitly enabled with safe_load

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - Remove confusing README section about the pane-base-index and
   window-base-index options. These options can be set independently of one
   another now that #542 and #543 are merged.
+- Allow YAML Anchors & Aliases as per [spec](http://yaml.org/spec/1.2/spec.html#id2765878)
 
 ## 0.11.3
 ### Misc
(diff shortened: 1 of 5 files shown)


agios:
Also added a test.

@tmuxinator/tmuxinator has anyone had a chance to look at this? YAML aliases allow one to have global configuration default template snippets.

ethagnawl:
Thanks for this, @agios. I will try to review it today.

ethagnawl:
@agios I just experimented with this locally and it works well. This will make for a nice addition to tmuxinator!

Would you mind adding an entry to CHANGELOG.md using the instructions in CONTRIBUTING.md?

agios:
Added changelog note. This used to work already before, it was inadvertently broken by https://github.com/tmuxinator/tmuxinator/commit/2cd3c675feaf02967633b986f16e6be8b928b003.

Perhaps what might be more interesting is a use case for this. Eg, you can use a defaults file, with the structure for various types of projects (rails, rails+docker, etc) and in an actual project file just write:

```yaml
name: test
root: ~/test

<%= Erubis::Eruby.new(File.read("#{ENV['HOME']}/.tmuxinator/defaults")).evaluate(self) %>

<<: *rails-docker-defaults
```

PR #590, opened 2018-02-07
Fix error when using wemux instead of tmux

Flink:
Currently the following error is thrown when trying to use wemux instead of tmux:
> undefined method `class_eval' for <#Tmuxinator::Project:0x00000002931528>

This patch extend the current instance of `Tmuxinator::Project` only if wemux is detected without using `class_eval` and avoids this error completely.

### DIFF
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -3,7 +3,6 @@ module Tmuxinator
     include Tmuxinator::Util
     include Tmuxinator::Deprecations
     include Tmuxinator::Hooks::Project
-    include Tmuxinator::WemuxSupport
 
     RBENVRVM_DEP_MSG = <<-M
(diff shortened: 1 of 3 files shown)


Flink:
Submitting this PR as a suggestion on how to resolve the current issue with wemux. I’m aware about https://github.com/tmuxinator/tmuxinator/pull/453 but it seems this PR is outdated and is a very big rewrite compared to mine.

Flink:
@ethagnawl Any chance for this PR to be reviewed/approved/merged or something? :sweat_smile:

ethagnawl:
Hey, @Flink. There's definitely a chance. Let's see if we can get someone with wemux experience to have a look at this PR. 

/cc @tmuxinator/tmuxinator

ethagnawl:
@Flink This looks good to me and, IMO, is an improvement over the previous approach.

Flink:
Thanks! :smile:

PR #598, opened 2018-03-17
Fix configuration location files listed in README

### LINKED ISSUE #597: Missed project configuration file at README
This is a minor, trivial issue with the documentation about the list of project configuration files. I think, [based on source code](https://github.com/tmuxinator/tmuxinator/blob/503d62b10e6c598fd2593628f9476d905a1f03fc/lib/tmuxinator/config.rb#L23), that there is missed `~/.config/tmuxinator` so I prepared a small PR. But following the [contribution guidelines](https://github.com/tmuxinator/tmuxinator/blob/master/CONTRIBUTING.md) I am creating this issue first.

Sorry in advance if I were wrong :)

dgdavid:
> Based on source code, specifically in comment at line
https://github.com/tmuxinator/tmuxinator/blob/503d62b10e6c598fd2593628f9476d905a1f03fc/lib/tmuxinator/config.rb#L23
> Resolve #597

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,8 @@ Tmuxinator will attempt to use the following environment variables (in this
 order) when creating or searching for existing project configuration files:
 
 - `$TMUXINATOR_CONFIG`
-- `$XDG_CONFIG_HOME/tmuxinator`
+- `$XDG_CONFIG_DIR/tmuxinator`
+- `~/.config/tmuxinator`
 - `~/.tmuxinator`
 
 This behavior is opt-in. If you do nothing, Tmuxinator will continue to use


ethagnawl:
Thanks for this, @dgdavid.

I just looked over the config class, reviewed the XDG spec and ran some tests on my machine. I _think_ the comment you've referenced (`Is ~/.config/tmuxinator unless $XDG_CONFIG_DIR is set`) is incorrect. I see no mention of `XDG_CONFIG_DIR` (singular) in the spec and setting `XDG_CONFIG_DIR` seems to have no affect on Tmuxinator. 

I think the README should be left as is and the comment should be changed to say something like, `Is ~/.config/tmuxinator unless $XDG_CONFIG_HOME is using a non-standard path`.

Any thoughts on the above, @HaleTom?

dgdavid:
> XDG_CONFIG_DIR (singular) in the spec and setting XDG_CONFIG_DIR seems to have no affect on Tmuxinator.

I see, you are right.

>  and the comment should be changed to say something like, Is ~/.config/tmuxinator unless $XDG_CONFIG_HOME is using a non-standard path

So, the change at README should be something like

```diff 
 -`$XDG_CONFIG_HOME/tmuxinator`
+- `~/.config/tmuxinator`
 - `~/.tmuxinator`
```

Isn't it?

ethagnawl:
@dgdavid IMO, the README should stay as it is -- on Master. I think using the variable makes sense because someone may be using a non-standard value for `$XDG_CONFIG_HOME` (e.g. `~/my-config`), in which case tmuxinator _will not_ attempt to use `~/.config`.

dgdavid:
>  someone may be using a **non-standard value for $XDG_CONFIG_HOME**

Hmmmm. Thanks for your reply, because now I understand this better :sweat_smile: 

Please, let me explain what happened.

Before I made the PR yesterday, I check the value of my `$XDG_CONFIG_HOME` first, which is none

```bash
❯ echo $XDG_CONFIG_HOME


```

Then, after looking for in my file system the recent created tmux session configuration file I found it under ~/.config/tmuxinator. Finally, I was taking a look to the source code and, although I did not find the explicit reference to this location I saw the quoted comment and, sorry for that, I did not investigate much more.

Now, thanks to your comments, I just review the [XDG documentation in the Arch wiki](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support#User_directories) where  is quite clear: **`~/.config/` is the default value for `$XDG_CONFIG_HOME`**

So, my apologizes for make you lose your time due to my lack of devops knowledge and not enough investigation :sweat_smile::sweat_smile::sweat_smile::sweat_smile:

I am closing the PR and related issue.

Again thank you so much.

Issue #595, opened 2018-03-02
Option to feed config file directly to tmuxinator on command-line.


SuperElectric:
This was originally a question on StackOverflow, to which the answer was "you can't do that with tmuxinator, but you could file a feature request": https://stackoverflow.com/q/49058693/399397

I have a tmuxinator config file (say, my_config.yaml,) stored in a source tree of a shared project.

For others to use it, AFAICT they first have to make a soft-link to it under their home directory, then launch tmuxinator separately:

```
$ ln -rs my_config.yaml ~/.tmuxinator/
$ tmuxinator my_config
```
This is inconvenient, and also requires that all your config files have unique filenames, lest they collide in ~/.tmuxinator/. I could use project root config file directories, but this still requires a link if I want the script to live under a different directory than the project root.

I'd much rather just be able to do something like:

```
$ tmuxinator my_config.yaml
```

Could there be way to pass the config file to tmuxinator on the command line?

Hamitamaru:
I'd also like this feature.

I'm in a weird place that tmuxp has this ability but does not have the same ERB functionality in their config files that are supported by tmuxinator.

If tmuxinator added this feature I can move over.

ethagnawl:
Mirroring my answer to @SuperElectric's Stack Overflow question:

Tmuxinator does not currently support a config file flag. However, there is support for project level config files. If you move my_config.yml to /project/root/.tmuxinator, you can then start it (from the project root) via `tmuxinator start .` or `tmuxinator local`.

ethagnawl:
@SuperElectric @Hamitamaru This feature is contained in v0.11.0.

SuperElectric:
Awesome; thank you!

PR #601, opened 2018-03-19
Add project config file option to `tmuxinator start`

ethagnawl:
Users can now provide a project config file to `tmuxinator start` using the `--project-config` or `-p` flags. (e.g. `tmuxinator start -p my-project.yml`).

Addresses #595.

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -164,9 +164,9 @@ module Tmuxinator
       end
 
       def create_project(project_options = {})
-        attach_opt = project_options[:attach]
-        attach = !attach_opt.nil? && attach_opt ? true : false
-        detach = !attach_opt.nil? && !attach_opt ? true : false
+        attach_opt = project_options.fetch(:attach) { false }
+        attach = attach_opt
+        detach = !attach_opt
 
         options = {
           args: project_options[:args],
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -272,12 +272,12 @@ module Tmuxinator
       new_config_path = Tmuxinator::Config.project(new)
 
       exit!("Project #{existing} doesn't exist!") \
-        unless Tmuxinator::Config.exists?(existing)
+        unless Tmuxinator::Config.exists?(name: existing)
 
-      new_exists = Tmuxinator::Config.exists?(new)
+      new_exists = Tmuxinator::Config.exists?(name: new)
       question = "#{new} already exists, would you like to overwrite it?"
       if !new_exists || yes?(question, :red)
-        say "Overwriting #{new}" if Tmuxinator::Config.exists?(new)
+        say "Overwriting #{new}" if Tmuxinator::Config.exists?(name: new)
         FileUtils.copy_file(existing_config_path, new_config_path)
       end
 
@@ -290,7 +290,7 @@ module Tmuxinator
 
     def delete(*projects)
       projects.each do |project|
-        if Tmuxinator::Config.exists?(project)
+        if Tmuxinator::Config.exists?(name: project)
           config = Tmuxinator::Config.project(project)
 
           if yes?("Are you sure you want to delete #{project}?(y/n)", :red)
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -169,11 +169,12 @@ module Tmuxinator
         detach = !attach_opt.nil? && !attach_opt ? true : false
 
         options = {
+          args: project_options[:args],
+          custom_name: project_options[:custom_name],
           force_attach: attach,
           force_detach: detach,
           name: project_options[:name],
-          custom_name: project_options[:custom_name],
-          args: project_options[:args]
+          yaml_filepath: project_options[:yaml_filepath]
         }
 
         begin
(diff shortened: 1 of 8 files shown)


Hamitamaru:
Would it be possible to do:
`tmuxinator start -p config-for-tmux-session-1.yml -p config-for-tmux-session-2.yml ...`

Then when tmux opens I can just switch between the sessions.

lib/tmuxinator/cli.rb
```diff
@@ -164,16 +164,17 @@ def generate_project_file(name, path)
       end
 
       def create_project(project_options = {})
-        attach_opt = project_options[:attach]
-        attach = !attach_opt.nil? && attach_opt ? true : false
-        detach = !attach_opt.nil? && !attach_opt ? true : false
+        attach = detach = false
+        attach = true if project_options[:attach] == true
+        detach = true if project_options[:attach] == false
```
adamstrickland: Would this be better as

```ruby
attach = project_options[:attach]  # will be a boolean-ish string value
detach = !attach # will be an actual TrueClass/FalseClass
```

lib/tmuxinator/config.rb
```diff
@@ -123,20 +126,43 @@ def directories
         end
       end
 
+      def valid_project_config?(project_config)
```
adamstrickland: Totally not relevant to this PR: we should prolly think about moving these validations to a separate file/module.  Getting a bit crowded in here...  Again, not germane to this...  

Nothing to see here

ethagnawl:
@Hamitamaru That is definitely possible. I'd prefer to explore that option separately, though. Feel free to open a new (feature request) issue.

lib/tmuxinator/config.rb
```diff
@@ -123,20 +126,43 @@ def directories
...
       def validate(options = {})
         name = options[:name]
         options[:force_attach] ||= false
         options[:force_detach] ||= false
-
-        project_file = if name.nil?
-                         raise NO_LOCAL_FILE_MSG \
-                           unless Tmuxinator::Config.local?
+        project_config = options.fetch(:project_config) { false }
+        project_file = if valid_project_config?(project_config)
```
adamstrickland: If I'm reading this right, `valid_project_config?` may raise an error, which is uncaught here and therefore needs to be handled by the caller of `validate`.  I don't really like that pattern, TBH, although we are [essentially already using it](https://github.com/tmuxinator/tmuxinator/blob/master/lib/tmuxinator/config.rb#L132)...  However, I would prefer if the error-raising was consolidated into one place, rather than distributed among multiple methods.  Maybe

```ruby
def valid_standard_project?(name)
  name ? true : false
end

def validate(options = {})
  valid = valid_project_config?(project_config) || valid_local_project?(name) || valid_standard_project?(name)

  if valid
    # now check/handle exception conditions
    raise "Project #{name} doesn't exist." unless exists?(name: name)
  
    # do other stuff?
  end
end
```

Thoughts?

spec/lib/tmuxinator/config_spec.rb
```diff
@@ -329,6 +329,37 @@
   describe "#validate" do
     let(:default) { Tmuxinator::Config::LOCAL_DEFAULT }
 
+    context "when a project config file is provided" do
```
adamstrickland: If we're going to keep the existing implementation (`raise`ing errors in the validations) we should test for that behavior.

lib/tmuxinator/cli.rb
ethagnawl: This piece of code is more confusing that it should be because of how the `attach` option is overloaded. The `attach`/`detach` variables are mutually exclusive iff `project_options[:attach]` is not `nil`. Also, AFAIK, Thor coerces "f" and "false" to `false` and any other string or the empty flag to `true`, so I don't think there's any reason to cast. I'll add a comment which documents this.

FWIW, I had actually used your approach in my initial refactoring (which was prompted by Code Climate failing the build -- and the only reason I touched any of this) but realized that it didn't cover all of possible scenarios. There were also no tests covering these assignments, which is why I added them.

I'm thinking that we should introduce a top-level detach option. That would remove all of the conditional logic and simplify this method. I'll open a separate issue which addresses the possibility of adding the new option.

lib/tmuxinator/config.rb
ethagnawl: Yeah, I agree that those raises in the helper methods are less than Ideal. I had introduced the new conditional/raise alongside the others in the containing method and was going to open an issue in order to explore creating classes/structs to encapsulate the different project types and their validation behavior. However, in the meantime, Code Climate failed the build as a result of the new conditional and the simplest way forward was to move that validation logic into a method for each scenario.

I'll noodle around some more with your suggestion in mind and see if I can come up with a more clear, simpler solution.

adamstrickland: Gotcha.  Gotta love how CC sometimes pushes us to less-than-ideal solutions over something like that :)

If we think we're ok via test coverage for now, I'd say :shipit: and let's take a look at figuring out a better long-term solution.

adamstrickland:
@ethagnawl This seems to me to be something we should ship a minor-version release for, agree?

ethagnawl:
@adamstrickland Agreed on the version bump. I'll merge #604, pull/rebase and update CHANGELOG.md later this evening.

Shall we sit on this for another few days to see if anyone else has feedback? Also, thanks for the thorough review!

ethagnawl:
@Hamitamaru Just a reminder to create that feature request for multiple configs, if that's something you're still interested in.

@adamstrickland I've created issues for the bits we discussed above and a few others (#606, #607, #608).

Do you want to wait a while longer for feedback from others or would you like to do the honors?

adamstrickland:
:shipit:

PR #603, opened 2018-03-22
Always prefer local lib over other versions

tribut:
`>>` appends (to `LOAD_PATH`) instead of prepending, which means another installation might take precedence. This bug was introduced in 2903c70e.

With the current code, this happens for me:

~~~~
$ $GEM_HOME/gems/tmuxinator-0.10.1/bin/tmuxinator version
tmuxinator 0.8.1
~~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## Unreleased
+- Fix load order when multiple versions of tmuxinator are installed (#603)
+
 ## 0.11.0
 ### Misc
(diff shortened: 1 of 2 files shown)


ethagnawl:
Thanks for this, @tribut! 

Per the [Contributing Guidelines](https://github.com/tmuxinator/tmuxinator/blob/master/CONTRIBUTING.md#pull-requests), would you mind updating CHANGELOG.md with a summary of the fix?

Otherwise, this looks good to me. :+1:

tribut:
Sure, sorry for that.

PR #610, opened 2018-04-05
Fix zsh completion

ZeroKnight:
## The Problem

If the user has no Tmuxinator projects, then this line:

```zsh
_describe -t projects "tmuxinator projects" projects
```

Will cause this chaos:

badcompletion

Notice how the command descriptions are actual selectable completions, the random gaps, and, while not pictured, all output repeats a couple of times (notice the 126 total matches at the bottom).

## The Fix

Very minor change: If the `projects` array is empty, don't try to run the `_describe -t projects ...` line. This change fixes the erratic results in the above screenshot with the settings listed below.

Also, it's better not to run `_describe` if there's no need to, anyway. 😃

## Minimal Reproduction

With a brand new user, with shell set to `zsh`, and **no** Tmuxinator projects:

**.zshrc**

```zsh
autoload -Uz compinit && compinit
source '<path/to/tmuxinator.zsh>'

zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|[._-]=* r:|=*'
```

Type 'mux' or 'tmuxinator', followed by a space, and then press `Tab`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - Remove confusing README section about the pane-base-index and
   window-base-index options. These options can be set independently of one
   another now that #542 and #543 are merged.
+- Fix zsh completion when there are no projects
 
 ## 0.11.3
 ### Misc
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -4,8 +4,9 @@ _tmuxinator() {
   projects=(${(f)"$(tmuxinator completions start)"})
 
   if (( CURRENT == 2 )); then
-    _describe -t commands "tmuxinator subcommands" commands
-    _describe -t projects "tmuxinator projects" projects
+    _alternative \
+      'commands:: _describe -t commands "tmuxinator subcommands" commands' \
+      'projects:: _describe -t projects "tmuxinator projects" projects'
   elif (( CURRENT == 3)); then
     case $words[2] in
       copy|debug|delete|open|start)


ethagnawl:
@tmuxinator/tmuxinator Is anyone knowledgeable enough about zsh to address this PR?

adamstrickland:
I am not

ethagnawl:
Since this seems like a pretty benign change and since @ZeroKnight seems to know what they're doing with regard to zsh (I had a quick look at their dotfiles), I'm okay with merging it as-is. 

Thoughts?

ZeroKnight:
I'm busy with finals currently, but later on today or tomorrow I could see
if the knowledgeable folk in the Zsh mailing list and/or #zsh on Freenode
would review it, if it would provide some peace of mind 🙂

ZeroKnight:
So someone from #zsh was kind enough to review my change, and offered a much nicer change:

```
17:49:25 <okdana> that will probably work
17:49:37 <okdana> you could also use `_describe -x` or _message
17:49:59 <okdana> the description is supposed to be singular tho
17:57:37 <okdana> does _describe give an error if the array is empty? mine doesn't seem to
17:57:53 <okdana> i know _values does but it takes them as separate operands
18:11:25 <ZeroKnight> No error, but there's the completion output gets really strange like in the screenshot
18:11:43 <ZeroKnight> The key to making it happen, and I don't know why, is the matcher zstyle in the minimal reproduction
18:12:04 <ZeroKnight> zsh 5.4.2 on Arch Linux, if that helps
18:18:30 <okdana> oh i see what you mean
18:25:39 <okdana> https://ptpb.pw/60b9.M17byPE6
18:25:43 <okdana> it doesn't seem to break if you do it like that
18:26:42 <okdana> and you won't get return status 1 if your projects are missing
```

I've implemented that change as a new commit to this PR. It works just as my original commit does, but this, I feel, is much cleaner. I can squash it to one commit using either implementation, just let me know what you'd like to do.

ethagnawl:
Thanks for revisiting this, @ZeroKnight! 

If you wouldn't mind updating the CHANGELOG (see the instructions in the README), we can get this merged today.

ZeroKnight:
Okay, added a line to the Changelog and squashed to one commit. I realize that this PR wasn't done on a topic branch; I would switch it to one, but I'm not sure how to do it without opening a new PR here.

Issue #555, opened 2017-09-20
'on_project_stop' hook with docker-compose


iovis9:
I'm trying the following to start/stop my docker-compose configuration:
```yaml
name: some_site
root: ~/Sites/some_site

on_project_first_start: docker-compose up -d
on_project_stop: docker-compose stop

windows:
  - editor: $EDITOR -S ~/.dotfiles/.vim/sessions/some_site.vim
```

`on_project_first_start` works as expected, launches the docker-compose configuration nicely! But when I stop the tmux session (tried exiting all terminals inside, running `:kill-session` and killing the session when detached) the docker server is still running. I don't know how to debug this or when is this hook supposed to run.

ethagnawl:
Hey, @iovis9.

It looks like this hook is only run when you use the `tmuxinator stop some_site` command. (This feature needs better documentation and I'll open a separate issue for that.)

However, I was just reviewing the pull request this feature came in on and I mentioned that I'd had trouble with that command on my system (Debian 8/9) back then. Are you running Linux, too?

iovis9:
I'm running the latest macOS. I just tried `tmuxinator stop some_site` with the following:
1. Detaching first and then running the command: working
2. Running the command from the actual session: kills the session correctly but doesn't fire up the hook

ethagnawl:
Ping @endorama. 

Do you have any idea what's going on here? Is this the expected behavior?

endorama:
Hello! I'm on holiday right now and I can't debug it properly, sorry. I'll
have a better look next week when I'm back!
>
>
For what I remember, this is not expected to work just killing tmux,
because of tmuxinator behaviour: the `on_stop` hook is run when the command
`tmuxinator stop` is run, not when the tmux session/server is killed.

Also, running tmuxinator stop from inside tmux will probably not work as
the tmux session ( and the shell within ) is killed before running the
`on_stop` command ( look at lib/tmuxinator/assets/template-stop.erb ).

As tmuxinator just create and run a shell script for performing it's
actions, to debug it's behaviour you can print on screen the command before
it's execution ( temporary hacking a bit the source code of the gem ).

The expected behaviour is: run the command after killing the tmux session.

I remember reading about tmux hook that could be used for this, but I was
not proficient enough to do it. If someone could help, I'll gladly have a
look again.

@ethagnawl if you want to assign the documentation PR to me highlighting
what need clarification I'll submit a PR.

@iovis9 if you still need help let me get back from holidays and I'll help
[...]

PR #617, opened 2018-05-07
Fix on exit not firing

endorama:
This PR should address #555, restoring `post`-like behaviour for `on_project_exit`.

That behaviour was prevented by the `unless attach?` check, which AFAICT is not useful in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+- Fix [#555](https://github.com/tmuxinator/tmuxinator/issues/555), restoring
+  `on_project_exit` hook behaviour ( same as deprecated `post` )
+
 ## 0.11.1
 ### Misc
 - Add support for tmux 2.7 (#611)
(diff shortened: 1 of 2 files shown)


ethagnawl:
@endorama It's been a while since I looked at the linked issue. Might people have come to rely on the existing/"broken" behavior or is this only a bug fix?

codezomb:
@ethagnawl the linked issue is requesting a bug fix. The issue describes that the current documented functionality does not work at all.

endorama:
@ethagnawl AFAICT that functionality never worked as intended, and I don't think can be used like this.

But if you can provide some more informations on the `attach?` method this should be clearer.

My intent with the `unless attach?` check was to avoid that multiple shell connected to the same tmux instance would each trigger the `on_project_exit` hook on session detach. With that check instead the `on_project_exit` is never triggered, and I would consider this a bug.

ethagnawl:
I just reviewed the linked thread and tested this locally. I can confirm that everything appears to be working as expected: `on_project_exit` is run (_once_) on `tmux detach`, `C-d` and `tmux kill-session ...`.

This fix gets a :+1: from me.

Any thoughts, @adamstrickland?

Issue #431, opened 2016-06-28
`.tmuxinator.yml` in ./ overrides mux basic functionality


hholst80:
If you put a `.tmuxinator.yml` in the current directory you cannot use `mux help` or other commands instead tmuxinator auto-starts and/or attaches to the session defined in the local config.

version 0.7.0 (in Ubuntu 16.04 LTS)

podung:
Confirmed on 0.7.1 also

podung:
@adamstrickland thoughts?

J3RN:
I think this should actually be pretty trivial. I'll revisit the logic in `/bin/tmuxinator` and see what goes on in there. Thanks for reporting!

J3RN:
Oh gosh, I just realized that this is a v0.7.* issue. This was probably resolved in the v0.8.0 release. @hholst80 Is there any reason you haven't updated to a later version?

ethagnawl:
I'm still seeing this on 0.8.1.

![local-before](https://cloud.githubusercontent.com/assets/289949/17010425/d4b4459c-4ed1-11e6-9fd0-5556f190f0d2.png)
![local_after](https://cloud.githubusercontent.com/assets/289949/17010424/d4b29652-4ed1-11e6-8eb4-633eb4df1ea8.png)

adamstrickland:
🐛

jpaugh:
Still seeing this in `0.10.0`. However, some commands *do* work.

These do not work:

    $ mux -v
    $ mux help

These commands do work:

* `version`
* `list`
* `commands`

I haven't tried any other commands, because I haven't learned what they do, yet.

adamstrickland:
Crap...  I self-assigned this some time ago and totally flaked.  Sorry about that.

jpaugh:
Life happens. That's okay. 

On December 7, 2017 3:52:53 PM CST, Adam Strickland <notifications@github.com> wrote:
>Crap...  I self-assigned this some time ago and totally flaked.  Sorry
>about that.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/tmuxinator/tmuxinator/issues/431#issuecomment-350105703

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #628, opened 2018-05-27
Local config overrides basic commands

adamstrickland:
FIXES https://github.com/tmuxinator/tmuxinator/issues/431

## Problem

Branching logic in `bin/tmuxinator` caused commands that would normally fall through to the base `thor` implementation (e.g. `-v`, `help`) to be consumed by the `tmuxinator` startup, which produced some weird behavior.

## Solution

- Moves branching logic to `Tmuxinator::Cli.bootstrap`, to enable testing.
- Refactors command lookup:
  - Creates `RESERVED_COMMANDS`, which is the superset of `tmuxinator` commands (i.e. existing functionality) and `thor` commands (at least, those explicitly denoted in the issue).

## Next Steps

While this is a decent stopgap, IMO the CLI implementation is getting pretty convoluted.  I'm thinking that perhaps I will spike on a better long-term solution...  Prolly splitting the CLI into 3 or 4 parts:

- Base functionality (`list`, `commands`, etc)
- Project functionality using centralized config location
- Project functionality using local config location (i.e. `--local`)
- Wemux functionality

Or something along those lines... I think it would decrease complexity and facilitate an easier debugging experience.  Opinions welcome.

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -345,5 +345,17 @@ module Tmuxinator
       say "Checking if $SHELL is set ==> "
       yes_no Tmuxinator::Doctor.shell?
     end
+
+    def self.bootstrap(*args)
+      name = args[0] || nil
+      if args.empty? && Tmuxinator::Config.local?
+        Tmuxinator::Cli.new.local
+      elsif name && !Tmuxinator::Cli::RESERVED_COMMANDS.include?(name) &&
+            Tmuxinator::Config.exists?(name: name)
+        Tmuxinator::Cli.new.start(name, *args.drop(1))
+      else
+        Tmuxinator::Cli.start(*args)
+      end
+    end
   end
 end
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -364,7 +364,7 @@ module Tmuxinator
     # array or ARGV, not a varargs.  Perhaps ::bootstrap should as well?
     # - ::start has a different purpose from #start and hence a different
     # signature
-    def self.bootstrap(*args)
+    def self.bootstrap(args = [])
       name = args[0] || nil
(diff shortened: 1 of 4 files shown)


lib/tmuxinator/cli.rb
```diff
@@ -344,5 +355,25 @@ def doctor
...
+
+    # This method was defined as something of a workaround...  Previously
+    # the conditional contained within was in the executable (i.e.
+    # bin/tmuxinator).  It has been moved here so as to be testable. A couple
+    # of notes:
+    # - ::start (defined in Thor::Base) expects the first argument to be an
+    # array or ARGV, not a varargs.  Perhaps ::bootstrap should as well?
+    # - ::start has a different purpose from #start and hence a different
+    # signature
+    def self.bootstrap(*args)
```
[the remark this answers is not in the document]
ethagnawl: I'd vote to continue using `::bootstrap` as is.

lib/tmuxinator/cli.rb
```diff
@@ -344,5 +355,25 @@ def doctor
       say "Checking if $SHELL is set ==> "
       yes_no Tmuxinator::Doctor.shell?
     end
+
+    # This method was defined as something of a workaround...  Previously
+    # the conditional contained within was in the executable (i.e.
+    # bin/tmuxinator).  It has been moved here so as to be testable. A couple
+    # of notes:
+    # - ::start (defined in Thor::Base) expects the first argument to be an
+    # array or ARGV, not a varargs.  Perhaps ::bootstrap should as well?
```
ethagnawl: Using a regular parameter does seem cleaner and, unless I'm missing something, I can't think of a scenario where the splat would be useful. Seems like after removing the two `*`s, everything should continue to work? If not, now it'll be easy enough to test. ;)

adamstrickland: Done

Issue #629, opened 2018-05-29
Layout option / panes not being sent through to tmux


JonRowe:
Using the latest version of tmuxinator `0.11.2`, panes and layout for a window doesn't do anything, project file:

```
project_name: td
project_root: ~/Code/TheProject/Dashboard
windows:
  - editor: vim
  - shell:
    layout: main-vertical
    panes:
      - git status
      - yarn run
```

Output:
```
#!/bin/bash

# Clear rbenv variables before starting tmux
unset RBENV_VERSION
unset RBENV_DIR

tmux start-server;

cd /Users/jon/Code/TheProject/Dashboard

# Run on_project_start command.


  # Run on_project_restart command.
  

  if [ -z "$TMUX" ]; then
    tmux -u attach-session -t td
  else
    tmux -u switch-client -t td
  fi


    
# Run on_project_exit command.
```

JonRowe:
Ah yes, the whitespace was the issue, would you accept a PR noting this on the read me?

ethagnawl:
So, @JonRowe. I've determined this issue is actually YAML's fault. It's a quirk in how it handles the indentation of list elements. Specifically, where the indentation actually begins. I've yet to find a definitive source (e.g. in the YAML spec) but I have found references to this issue in a few places.

> Notice that while the indentation is two spaces per level, for the values under the context and defaults options there is a four-space indent. If only two spaces are used to indent, then those keys will be considered part of the same dictionary that contains the context key, and so the data will not be loaded correctly.

https://docs.saltstack.com/en/2017.7/topics/troubleshooting/yaml_idiosyncrasies.html

> Block list items can (but don't need to) have the same indentation as the surrounding block level because "-" is considered part of the indentation

https://stackoverflow.com/a/42248984/382982

[My own SO question about your use case.](https://stackoverflow.com/questions/50594758/why-isnt-two-spaced-yaml-parsed-like-four-spaced-yaml)

I'm looking forward to seeing how you address this is your PR. Thanks, again, for the offer!

ethagnawl:
Closed via #630.

PR #630, opened 2018-05-30
Highlight spacing issue.

JonRowe:
This is an attempt to highlight the fact that spacing is a bit funky for YAML, see #629 for background.


JonRowe:
Looks good to me

Issue #265, opened 2014-11-15
Option to create window in the same tmux session instead of creating a new session


pengux:
Most of the time for my workflow, it's enough just to create a new window with multiple panes in it instead of creating new sessions.

olalonde:
Same problem here. If I'm working on multiple projects, I like to have one window per project and move from one window to the other. So I'd like tmuxinator to be able to start multiple projects in the same session.

tobixx:
Just want to raise my hand for this issue too!

phcerdan:
I want to raise my hand too.
To simplify and avoid going into projects inside projects, and probably nested tmux sessions.
Should be possible to load one, and only one, window layout in the current window? Similar to tmux select-layout, but adding the feature of loading defined scripts in each pane.

It would provide modularity, and run time access to 1 window projects. And it would be great!

olalonde:
I ended up writing my tiny alternative to tmuxinator in <100 lines of bash: https://github.com/olalonde/muxide

It's not configurable at the moment but the script is pretty tiny. It launches in the current window if there are no other panes in the window or in a new window if there are panes in current window or in a new tmux session + window if launched from outside tmux.

I plan to eventually make it configurable through a `.muxide` file in project roots.

assembler:
:+1:

scottwio:
Same request here
    mux append project1
would be prefect

cskeppstedt:
:+1:

skycaptain:
What's the current status of this issue?

pwnmagic:
+1

J3RN:
I'll look into this today and determine how long this will take to complete. However, if someone wants to implement this and send a pull request... 🌈

stedwick:
👍

flovilmart:
I actually liked a lot this feature and attempted to implement it.

You can try it from my branch with:

``` 
git clone https://github.com/flovilmart/tmuxinator
cd tmuxinator
gem build tmuxinator.gemspec
gem install tmuxinator-0.12.0.gem
```

stedwick:
Cool!

MateuszKubuszok:
Since this PR is stack for a while, I might share a workaround for my common use case, that is having open tmux session where I want to drop it and replace it by a new session from tmuxinator without leaving tmux

```
function tmux-into() { tmuxinator start $1 --no-attatch && tmux switch -t $1 && exit }
```

lougreenwood:
[...]

PR #652, opened 2018-09-16
Adds support for tmuxinator start --append

flovilmart:
Closes: #265

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,9 +11,10 @@ jobs:
       fail-fast: false
       matrix:
         ruby_version:
-          - '3.0.1'
-          - '2.7.3'
-          - '2.6.7'
+          - '3.3.0'
+          - '3.2.3'
+          - '3.1.4'
+          - '3.0.6'
         tmux_version:
           - '3.4'
           - '3.3a'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Replace exists? with exist? for Ruby >= 3.2
 ### Misc
 - Remove `mux` alias from Fish and Bash completions
+- Add Ruby 3.2, 3.1 to the test matrix; rm ruby < 3>
 
 ## 3.1.2
 ### Fixes


ethagnawl:
Thanks for this, @flovilmart! I'll try to give it a proper review before the end of the week.

DestyNova:
Very nice. The only confusing bit is that it names the new window `<session-name>:<window-name>` -- i.e. if I append a window called `muffin` from a file containing the line `name: mountain`, then the new window will be called `mountain:muffin`. If I don't specify the `name` parameter in the Yaml file, then the window will be called `:muffin`.

I see there is some logic to control this behaviour, but I don't fully understand it. Is there an extra flag or configuration value needed to just have the window be called `muffin`?

flovilmart:
@DestyNova that would be a possible change, if the name is not set with the tmux template I could probably remove the separator. This was intended to make it easy to see which windows were appended.

DestyNova:
Yes I see the motivation for it and it makes sense -- in my case though, I expect to end up with a bunch of files that each describe one window with a couple of panes. Since I use windows and panes where other people might use sessions and windows, I don't need that namespacing. Maybe the behaviour could be specified with an optional flag (e.g. if a simple name is default, then add `-p` for `prefix-with-session-name`, or if the namespaced window name is default, then `-s` for `simple-window-name` or something)?

flovilmart:
I’m not sure I want to spend additional time adding more configuration flags to this PR. If the naming of the windows is problematic i’ll Remove it altogether and restore the original behavior instead of prefixing with the screen name.

DestyNova:
I like the PR either way and am really happy someone has finally done this. Rather than sidetrack the PR, I'll try to do a separate one adding such a flag afterwards. Thanks for doing this :)

ethagnawl:
@flovilmart This is looking pretty good!

I do have a few questions/thoughts/requests/etc. for you:

1.) Would you mind adding an entry to the "Other Commands" section of README.md in your PR which documents how this feature and its various CLI options are intended to be used?

The documentation will be useful for users going forward _and_ for people reviewing this PR. I wasn't following #265 closely and it took me a few minutes to figure out how this feature was intended to work.

2.) A missing/malformed `current_session: ...` seems to result in tmuxinator wantonly appending windows to whatever the most recently created tmux session is, which seems undesirable. Is this by design? Is there a way you can sniff out whether or not the `current_session` exists and either fail or start a new session if it doesn't? (I'll revisit the code myself, but I'm on my way out for the day.)

3.) Could you replace `puts` with `say` in Cli.rb?

4.) Perhaps the "Creating new session ..." message should contain the session name?

5.) We will need some tests around this new behavior. Did you ever get a chance to take a pass at writing any? I know you said you're not used to Ruby, so I can appreciate how that might be a bit daunting and it's awesome that you were able to put this together. (I don't think you're giving yourself enough credit!) If not, that's totally fine and I can take a pass at writing some and submit another PR to your branch.

flovilmart:
I Didn’t forget about this one, but my todo list keeps pushing it down. I hope to have more time soon

raiyuza:
@flovilmart Any ETA, we in 2019 already :)

flovilmart:
Well, yea, I need to but my head back into it :)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

icole:
Any issue with getting this merged now that the merge conflicts are resolved?

flovilmart:
@icole I need to work on the few tests and documentation. I haven't found a good time to do it. I'll try this weekend.

icole:
@flovilmart awesome, thanks for the response. If you could use any help with anything let me know. Would love to see this go out.

flovilmart:
Thanks @icole I'll do my best to address all the concerns and add the relevant tests. I'd love to get done with this one. Sorry to have dropped the ball.

flovilmart:
@icole Sorry for the late response. I changed a bit the logic so it uses `mux start --append [project]` instead of `mux append`. This simplifies a bit the logic.

I'll add some tests around the `in_current_session` logic to make sure this is readable easily and maintainable as well.

icole:
@flovilmart thanks so much! could we pull this in after you add those tests? I am dying to get this in since the version this is based off is older than what our projects Gemfile requires.

betoharres:
I was trying to figure out if tmuxinator had this feature; well, hope it gets merged!

flovilmart:
I completely forgot I had this one open, thanks for the ping 😅! I’ll do my best.

protex:
Any chance this gets merged? It would be super useful.

flovilmart:
@icole what what left for update? I wanna go ahead with this. Can you pin point in the code the different changes required for the merge?

ethagnawl:
From what I recall, it's a minor Code Climate nitpick. Unfortunately, the build is old enough that it's been dropped by CC. Could you try pushing a new commit in order to trigger a new build, @flovilmart?

flovilmart:
@ethagnawl the problem is that the `start` method now is too complex, not sure how to simplify it given it's already quite simple.

ethagnawl:
I haven't used Code Climate in a long time and don't remember how some of their metrics work. I'd say maybe removing the `project` variable and nesting those method calls would work, but the more likely/correct solutions would be extracting those conditionals or moving the has creation into its own method.

Does anyone else have any thoughts about this or the PR in general? CC/ @tmuxinator/tmuxinator

flovilmart:
@ethagnawl , finally, codeclimate passed! 🎉

ethagnawl:
Okay, that's great. I'll have to find some time to review this again. 

Sorry for the bother, but does anyone else have any thoughts? /CC @tmuxinator/tmuxinator

flovilmart:
thanks @ethagnawl !

lib/tmuxinator/config.rb
```diff
@@ -25,6 +25,10 @@ def home?
         File.directory?(home)
       end
 
+      def options_file
+        directory + ".yml"
```
ethagnawl, line 29: The "yaml" extension is also supported. I tried using a project config file with this extension and everything worked as expected. Does that imply that this method isn't being used or am I overlooking something?

lib/tmuxinator/project.rb
```diff
@@ -208,7 +217,12 @@ def tmux_options
     end
 
     def base_index
-      get_base_index.to_i
+      base = 0
+      if in_current_session
+        # Get the last window index + 1
+        base = 1 + `tmux list-windows | awk '{print $1}'`.split.last.to_i
```
ethagnawl, line 223: I think this awk logic should either be handled in Ruby code or we should add a method to the Doctor class which verifies that awk exists on the user's system.

lib/tmuxinator/cli.rb
```diff
@@ -226,6 +227,44 @@ def show_continuation_prompt
...
         Kernel.exec(project.kill)
       end
+
+      def tmux_session_exists
+        ENV["TMUX"] ? true : false
+      end
+
+      def should_append?(options)
+        options[:append] == true ||
+          (options[:append] != false && Tmuxinator::Config.options[:append])
```
ethagnawl, line 237: Could you walk me through the use case for the right side of this or?

kurktchiev:
any chance we can get this merged any time soon? it would really boost productivity with tmux/tmuxinator

lib/tmuxinator/cli.rb
flovilmart: the idea is that can set append in the `Tmuxinator::Config.options` so all sessions are appending by default.
if append is not provided by the `options` (case `!= false`) we check that it is set in the global Config.

lib/tmuxinator/config.rb
flovilmart: what do you suggest?

lib/tmuxinator/project.rb
flovilmart: fair, let's me see what I can do

cibinmathew:
@ethagnawl @flovilmart @icole Thanks for this work.
Can we get this merged or may be list down the blocking issues and open another to track the enhancements. We could pitch in to get this to a closure.

lib/tmuxinator/config.rb
ethagnawl: Either this method needs to support both extensions or, if I'm correct in that it's not being used, it should be removed.

ethagnawl:
> @ethagnawl @flovilmart @icole Thanks for this work.
> Can we get this merged or may be list down the blocking issues and open another to track the enhancements. We could pitch in to get this to a closure.

There are still some issues (noted above) which need to be worked through. (If you feel like running with what's been done and @flovilmart doesn't mind, feel free to pick them up.) It'd also be good to get feedback from others in @tmuxinator/tmuxinator.

flovilmart:
@ethagnawl let me get back on it

lib/tmuxinator/config.rb
```diff
@@ -56,6 +60,15 @@ def default
...
       end
 
+      def options
+        yaml = begin
+          YAML.safe_load(File.read(options_file), [], [], true)
+        rescue SyntaxError, StandardError
+          return {}
+        end
+        yaml ? yaml : {}
+      end
```
noahfrederick, line 70: I believe this is equivalent:

```suggestion
      def options
        YAML.safe_load(File.read(options_file), [], [], true) || {}
      rescue SyntaxError, StandardError
        {}
      end
```

lib/tmuxinator/cli.rb
```diff
@@ -226,6 +227,44 @@ def show_continuation_prompt
...
+      def append_in_session(options, name = nil)
+        has_tmux_session = nil
+        if should_append?(options)
+          has_tmux_session = tmux_session_exists
+          if !has_tmux_session
+            say "Creating session '#{name}'"
+          end
+        end
+        has_tmux_session
+      end
```
noahfrederick, line 249: This method is somewhat hard to follow because it's responsible for two things: (a) deciding whether a new session is created *and* (b) printing a message with the new session name. Given (a), it should be named `append_in_session?` or something like that, and always return a Boolean (right now it returns `true`, `false` or `nil`—the rest of the code doesn't appear to distinguish between the `false` and `nil` possibilities). But since it also has a side effect of printing a message and takes this extra `name` parameter, that wouldn't be right either. These things don't seem to belong in the same method.

In any case, it could be simplified to:

```suggestion
      def append_in_session(options, name = nil)
        if should_append?(options) && !tmux_session_exists
          say "Creating session '#{name}'"
          true
        end
      end
```

lib/tmuxinator/project.rb
noahfrederick, line 223: Tell tmux to only print the window indices:

```suggestion
        base = 1 + `tmux list-windows -F '#I'`.split.last.to_i
```

The output is like so:

```
1
2
3
```

lib/tmuxinator/window.rb
```diff
@@ -108,7 +108,8 @@ def tmux_window_command_prefix
     end
 
     def tmux_window_name_option
-      name ? "-n #{name}" : ""
+      project_name = project.in_current_session ? "#{project.full_name}:" : ""
+      name ? "-n #{project_name}#{name}" : ""
```
noahfrederick, line 112: I would prefer to drop this change, i.e., for the window to always be named without the prefix.

noahfrederick (commented):
Looking forward to this feature. I left a drive-by review of the code changes.

flovilmart:
Thanks @noahfrederick will have a look later

lib/tmuxinator/config.rb
```diff
@@ -56,6 +60,12 @@ def default
         "#{directory}/default.yml"
       end
 
+      def options
+        YAML.safe_load(File.read(options_file), [], [], true) || {}
```
ethagnawl, line 64: @flovilmart Would you mind using the new safe_load method signature (sans positional parameters two, three and four)? As-is, this will cause errors in some environments because of changes to Psych and libyaml.

It would look something like this: 

`YAML.safe_load(File.read(options_file), aliases: true)`

See [this](https://github.com/tmuxinator/tmuxinator/pull/823/files) PR for more context.

megalithic:
@flovilmart / @ethagnawl -- just pinging those that have worked on this. How are we feeling about your concerns being addressed and getting this merged?

flovilmart:
@megalithic I haven't had the chance to actually finish this one; yeah I know, that's kinda pathetic after all this time 🤣

ethagnawl:
@flovilmart I know your comment was in jest, but it's not pathetic at all. Your contributions are being provided at the expense of whatever else you might be doing with your life and are greatly appreciated.

megalithic:
@ethagnawl @flovilmart 💯 agreed! OSS is a gift! i'm grateful for each of you that contribute/create such amazing tools.

flovilmart:
Haha! Thanks y’all! very appreciated. I need to find some time for this. I see this is pretty requested.

ChrisYin:
I'm wondering when this patch can be merged? Looks like it has a long history from 2018.

martin-braun:
@flovilmart I know this issue probably hunts you in your dreams already, is there still any chance that you find the time to finish this?

akofink:
I attempted to rebase this with `tmuxinator:master`, but I accidentally pushed the latest master, which auto-closed the PR. I have an updated branch that I will use to re-open a new PR, since I can no longer push to this original branch. Apologies!

Issue #583, opened 2017-12-06
Option to quiet tmux version warning


adamstrickland:
As suggested in https://github.com/tmuxinator/tmuxinator/pull/574#issuecomment-349469851, add a `--quiet` flag/option to send tmux version warning(s) to /dev/null

p-becker:
Can be closed.

I apparently didn't use [one of the magic keywords](https://help.github.com/articles/closing-issues-using-keywords/) in the pull request 🙈

PR #659, opened 2018-11-18
Add quiet flag to prevent tmux version warning (#583)

p-becker:
This adds a `--quiet` flag for the `start`, `stop` and `local` commands, which will prevent the tmux unsupported version warning from being displayed. The short alias `-q` also works.

As mentioned in #583, this is useful for people running the tmux master version. While working on this, I noticed that the `stop` and `local` did not yet include the warning, even though they do interact with tmux. So I took the liberty of adding the warning and flag for those commands as well.

To make this change fairly easy and understandable I did some refactoring and rather small commits. As this is my first PR _ever_, I'd appreciate any and all feedback.

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -210,6 +210,10 @@ module Tmuxinator
         Kernel.exec(project.render)
       end
 
+      def version_warning?(suppress_flag)
+        !Tmuxinator::TmuxVersion.supported? && !suppress_flag
+      end
+
       def show_version_warning
         say Tmuxinator::TmuxVersion::UNSUPPORTED_VERSION_MSG, :red
         say
@@ -231,6 +235,8 @@ module Tmuxinator
                          desc: "Give the session a different name"
     method_option "project-config", aliases: "-p",
                                     desc: "Path to project config file"
+    method_option "suppress-tmux-version-warning",
+                  desc: "Don't show a warning for unsupported tmux versions"
 
     def start(name = nil, *args)
       # project-config takes precedence over a named project in the case that
@@ -248,7 +254,9 @@ module Tmuxinator
         project_config: options["project-config"]
       }
 
-      show_version_warning unless Tmuxinator::TmuxVersion.supported?
+      show_version_warning if version_warning?(
+        options["suppress-tmux-version-warning"]
+      )
 
       project = create_project(params)
       render_project(project)
@@ -256,13 +264,16 @@ module Tmuxinator
 
     desc "stop [PROJECT]", COMMANDS[:stop]
     map "st" => :stop
+    method_option "suppress-tmux-version-warning",
+                  desc: "Don't show a warning for unsupported tmux versions"
 
     def stop(name)
       params = {
         name: name
       }
-
-      show_version_warning unless Tmuxinator::TmuxVersion.supported?
+      show_version_warning if version_warning?(
+        options["suppress-tmux-version-warning"]
+      )
 
       project = create_project(params)
       kill_project(project)
@@ -270,9 +281,13 @@ module Tmuxinator
 
     desc "local", COMMANDS[:local]
     map "." => :local
+    method_option "suppress-tmux-version-warning",
+                  desc: "Don't show a warning for unsupported tmux versions"
 
     def local
-      show_version_warning unless Tmuxinator::TmuxVersion.supported?
+      show_version_warning if version_warning?(
+        options["suppress-tmux-version-warning"]
+      )
 
       render_project(create_project(attach: options[:attach]))
     end
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -210,6 +210,13 @@ module Tmuxinator
         Kernel.exec(project.render)
       end
 
+      def show_version_warning
+        say Tmuxinator::TmuxVersion::TMUX_MASTER_DEP_MSG, :red
+        say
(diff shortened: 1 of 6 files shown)


ethagnawl:
Thanks, @p-becker! I'll try to review this and get back to you with any feedback within the next few days.

lib/tmuxinator/tmux_version.rb
```diff
@@ -0,0 +1,29 @@
+module Tmuxinator
+  module TmuxVersion
```
ethagnawl: Perhaps this module name could be more intention revealing? (e.g. SupportedTmuxVersions)

lib/tmuxinator/cli.rb
```diff
@@ -210,6 +210,17 @@ def render_project(project)
         Kernel.exec(project.render)
       end
 
+      def version_warning?(quiet)
+        !Tmuxinator::TmuxVersion.supported? && !quiet
+      end
+
+      def show_version_warning
+        say Tmuxinator::TmuxVersion::UNSUPPORTED_VERSION_MSG, :red
+        say
```
ethagnawl: Is this second `say` necessary?

p-becker: This is just copy pasted from the deprecation warnings to keep the blank line between the message and the confirmation:
https://github.com/tmuxinator/tmuxinator/blob/8dab40337d85ee27ec27bf34cbdfb3290a0d5c72/lib/tmuxinator/cli.rb#L205
I guess it's easier to read this way?

But let me pull out the three duplicate lines, so that this would only need to change in one place.

p-becker:
I just realized that there is a problem with the method [self.bootstrap](https://github.com/tmuxinator/tmuxinator/blob/1b0776bf75ee21b7b38f5837d632d420fa5a491c/lib/tmuxinator/cli.rb#L412-L422) which lets you omit the `start` command, so you can type:
a) `tmuxinator [project_name]` instead of 
b) `tmuxinator start [project_name]`

In case a), `#start` gets called with all flags passed as args, instead of using the thor `options` hash. This means that none of the method_options will be used, including the `--quiet` flag.

Is this a separate issue or does this need to be handled here?

---
Some debug info:
**a)** 
```
rake install && tmuxinator dotfiles --name test --quiet --project-config ~/dotfiles/tmuxinator/dotfiles.yml --attach

"Hello from self.bootstrap elsif branch"
ARGS: ["--name", "test", "--quiet", "--project-config", "/Users/pbecker/dotfiles/tmuxinator/infra.yml", "--attach"]
OPTIONS: {}
    WARNING: You are running tmuxinator with an unsupported version of tmux.
    Please consider using a supported version:
    (1.5, 1.6, 1.7, 1.8, 1.9, 2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8)

Press ENTER to continue.
```

**b)**
```
rake install && tmuxinator start dotfiles --name test --quiet --project-config ~/dotfiles/tmuxinator/dotfiles.yml --attach
ARGS: []
OPTIONS: {"name"=>"test", "quiet"=>"quiet", "project-config"=>"/Users/pbecker/dotfiles/tmuxinator/dotfiles.yml", "attach"=>true}
```

ethagnawl:
Nice catch! I've created [a separate ticket for that issue.](https://github.com/tmuxinator/tmuxinator/issues/661) If you're up for it, you're more than welcome to make the fix and submit a PR.

adamstrickland:
I think I agree with @ethagnawl 's suggestion re: the option naming.  My preference would be to use `--suppress-tmux-version-warnings`.  It is very specific and self-documenting, and implies that the warning still exist, but the user wants simply to hide the output.  FWIW, I would also prefer *not* to give this option a shortened alias.

Here's a possible, smaller-footprint solution: reading the discussion that prompted this, the request was made simply so that

 > those using tmux master wouldn't have to hit 'enter' every time they fire up tmuxinator

So what if we fixed the problem, rather than worked around it?  The `enter` requirement is to simply stop execution and display a message before `tmux` takes over and wipes the current shell's output.  Perhaps it's enough to simply add the version warning into the `doctor` command and remove the interaction from the `start` mechanism.  Thoughts?

p-becker:
I agree that `--quiet` is too generic, unless it is really meant to silence everything.

I understand both the version and deprecations messages as warnings that things may break because you left the happy path. If that's the case, I think both kinds of messages should be treated equally.

That being said, how likely is the version to cause issues? If it's not very likely, let's hide it away in the `doctor` command.

adamstrickland:
We've supported every version of `tmux` since 1.5, which was released in 2011, so it's not very likely, IMO...  That said, leaving this as a runtime check (rather than a possibly-one-time-that-ran-when-tmuxinator-was-first-installed-three-years-ago check) is probably more in-line with the principle of least surprise.

OK, never mind re: the `mux doctor` thing.  Let's just change the `option` and :shipit:

p-becker:
I took the liberty to name it `...warning` instead of `...warnings` because there is only one version warning and we're trying to be precise :grimacing:

p-becker:
Two questions that may help to get this resolved:

- **Is this feature useful enough to justify the lines of code added?**
  - Playing devil's advocate here. I could totally understand if keeping the project's technical debt to a minimum outweighs adding features in some cases. Just because we can, doesn't mean we should.
- **Is this technically sound and ready to be merged?**

cc @ethagnawl @adamstrickland

ethagnawl:
Looks good to me. I was just waiting on the flag change.

PR #666, opened 2018-12-15
Fix failing spec with custom TMUXINATOR_CONFIG path

### LINKED ISSUE #665: Testing config_spec.rb fails with custom TMUXINATOR_CONFIG path
**Describe the bug**
When TMUXINATOR_CONFIG is a custom path, e.g. TMUXINATOR_CONFIG=/Users/_USERNAME_/dotfiles/tmuxinator, the `config_spec.rb` test will fail.

**To Reproduce**
`TMUXINATOR_CONFIG=$(pwd)/spec/fixtures/dot-tmuxinator bundle exec rspec spec`

**Expected behavior**
The spec does not fail.

**Environment**
 - tmuxinator Version: tmuxinator 0.13.0
- tmux Version: tmux next-2.9
 - Ruby Version: ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]
 - OS: mac OS mojave

p-becker:
Fix failing tests when TMUXINATOR_CONFIG is set

When the environment variable TMUXINATOR_CONFIG is set to a custom path,
the specs would fail*. This fixes #665 by putting a bandaid on the affected test.

I call it a bandaid because I am still not happy with the tests. Mostly because there is stubbing at many different levels:

- Entire methods
- Parts of methods, especially File.directory?
- Env variables

Stubbing ENV is not ideal, see for example
https://robots.thoughtbot.com/testing-and-environment-variables
But in this case, I get the feeling that setting up a cleanly stubbed
ENV would replace most of the other stubbing while being easier to
understand. This may be an idea for the future.

*Reproduce with:
TMUXINATOR_CONFIG=$(pwd)/spec/fixtures/dot-tmuxinator bundle exec rspec spec

### DIFF
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -9,9 +9,9 @@ module Tmuxinator
     class << self
       # The directory (created if needed) in which to store new projects
       def directory
-        return environment if File.directory?(environment)
-        return xdg if File.directory?(xdg)
-        return home if File.directory?(home)
+        return environment if environment?
+        return xdg if xdg?
+        return home if home?
         # No project directory specified or existant, default to XDG:
         FileUtils::mkdir_p(xdg)
         xdg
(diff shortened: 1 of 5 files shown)


ethagnawl:
This looks good to me. Anyone else have any thoughts? (/cc @tmuxinator/tmuxinator)

PR #672, opened 2019-01-03
671 - remove references to pre/post option

### LINKED ISSUE #671: Remove deprecated options from generated project config files
Deprecated project config options should not be present in generated project config files (ne sample.yml).

1. `pre`
2. `post`

ethagnawl:
Closes #671

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,10 @@ matrix:
   allow_failures:
     - env: TMUX_VERSION=1.7
 before_install:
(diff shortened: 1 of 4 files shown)


ethagnawl:
This PR also addresses an issue with the Travis config file. The build was broken because the config didn't specify/pessify? a bundler version number and, after bundler 2.0 was released on 1/3/2019, was installing bundler >= 2.0, which is incompatible with the version of bundler specified in tmuxinator.gemspec.

PR #674, opened 2019-01-04
673 remove support for ruby 2.2

### LINKED ISSUE #673: Drop support for Ruby 2.2
- Ruby 2.2 has been EOL'd
- Travis now uses Bundler 2.0 by default, which does not support Ruby 2.2, and the build is breaking as a result

ethagnawl:
Closes #673


tmuxinator.gemspec
```diff
@@ -37,15 +37,15 @@ Gem::Specification.new do |s|
...
+  s.required_ruby_version = ">= 2.3.6"
 
   s.add_dependency "erubis", "~> 2.6"
   s.add_dependency "thor", "~> 0.19", ">= 0.15.0"
   s.add_dependency "xdg", "~>2.2", ">= 2.2.3"
 
   s.add_development_dependency "activesupport", "< 5.0.0" # Please see issue #432
   s.add_development_dependency "awesome_print", "~> 1.2"
-  s.add_development_dependency "bundler", "~> 1.3"
+  s.add_development_dependency "bundler", "~> 2.0"
```
adamstrickland: out of curiosity, do we get anything in particular from upping this dependency version?

adamstrickland: rather: i understand removing the `~> 1.3`, which keeps us locked into the `1.x` versions of bundler, especially given that 2.0 has been released.  perhaps a better version constraint would simply be `>= 1.3`?

ethagnawl: Bundler 2.0 dropped support for Ruby 2.2, so I figured since we were doing the same we may as well follow suit. After some additional thought, though, I think you're right and we should opt for the less restrictive constraint.

Also, having read some more about Bundler 2.0, I think if we were to go with '~> 2.0', `required_rubygems_version` should be bumped since Bundler 2+ requires ~~3.0~~ [2.5](https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html).

Issue #678, opened 2019-01-31
Add newline list format flag


raiyuza:
**Is your feature request related to a problem? Please describe.**
I use a [custom build tmux](https://ershov.github.io/tmux/) which extends tmux configuration with a subset of the tcl language.

 I tried to get the project names from the list command but due an poor implementation of tcl or lack of understanding. I could only only get so so far without using quotes.  The furthest i got was something like the following.

```tcl
#choose from list and switch to project
bind-key P tcl {
  cd [f #{pane_current_path}]

  choose-from-list {*}[split [exec tmuxinator list | tail +2 | grep -oE {([0-9]|[aa-zz])\w+} | sort  ] "\n"] -onselect {
    exec git checkout [lindex [split [string range $_ 2 end] " "] 0]
}
```

All this work just to print the project names line by line, and even then i had to make additional iterations to get all the matches. 

This made me wonder why can't we do something like ls -1 for the list command.

**Describe the solution you'd like**
I suggest creating a flag called '--newline' to print each line like such :
```bash
$> tmuxinator list --newline
barfoo
foobar
yxz
```

This will make it easier and i assume more compatible with other text processing commands

**Describe alternatives you've considered**
```tcl
bind-key P tcl {
  cd [f #{pane_current_path}]

  choose-from-list {*}[split [exec sh findprojectslist.sh | sort  ] "\n"] -onselect {
    exec git checkout [lindex [split [string range $_ 2 end] " "] 0]
}
```
Writing my own PRQ

**Additional context**
This is my first issue ever created on github, any feedback is welcome!

PR #679, opened 2019-01-31
Add support list --newline (#678)

raiyuza:
Add support list --newline (#678)

created a --newline flag for the ``list`` command. Read up on issue #678 [here](https://github.com/tmuxinator/tmuxinator/issues/678)
```bash
-n, [--newline], [--no-newline]  # Force output to be one entry per line
```

Small sidenote:
this is my first public MRQ and ruby contribution, any feedback is welcome

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## UNRELEASED
+- Add `-n, [--newline], [--no-newline]` flag for list command
+    Force output to be one entry per line 
 ### Misc
 - make pre/post deprecation warnings more descriptive
 - remove pre/post from project configuration template
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -31,7 +31,7 @@ module Tmuxinator
       implode: "Deletes all tmuxinator projects",
       version: "Display installed tmuxinator version",
       doctor: "Look for problems in your configuration",
-      list: "Lists all tmuxinator projects"
+      list: "Lists all tmuxinator projects",
     }.freeze
 
     # For future reference: due to how tmuxinator currently consumes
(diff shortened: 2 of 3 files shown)


raiyuza:
Uploaded wrong code while trying to squash rookie mistake, hold on!

raiyuza:
1. I will in a few minutes! 

> 2. Maybe the newline option should become the default? Though, I guess this could potentially break other people's workflows in the opposite direction and could warrant a deprecation warning for a release cycle or two..

Hmmm,

We can keep the default behavior by giving it a seperator flag ala awk -F though. Users can set an env variable to control the default value of this flag. This way you don't have to break backwards compatibility and it makes the feature more useful overall.

In the same fashion it would be amazing to have have some kind of --format option, it would be able te
format how the list command would show your list, and extract properties from said projects. Ex.

```
$> mux l -F "\n" --format "@name\t@root\t@windows.editor.layout"
$> 
foo<tab>~/foo<tab>vertical
bar<tab>~/bar<tab>horizontal
```

This opens up an powerful way to create a formatted data structure. This mainly used for parsers in Ex.

 - vim, you could write a unite/denite script to read the outputs buffer to make menu's 
 - fuzzy search

Ill create an issue with this idea, better to discuss it there.

Thanks for the feedback!

ethagnawl:
Thoughts, @tmuxinator/tmuxinator?

rglassett:
Neat feature. Ideally I'd like to see a test assertion against the actual output format, but most of the other CLI tests don't really do this either, so the new test meets the current bar IMO.

:+1: from me.

raiyuza:
I created a new issue with the new idea [here](https://github.com/tmuxinator/tmuxinator/issues/681) .

Let's keep the scope of this issue/PRQ to the -n flag only. Makes it easier to segment what we are talking about!

Issue #563, opened 2017-09-27
Distribute Tmuxinator with Homebrew, Snap, and Flatpak


J3RN:
A large number of issues that users encounter whilst using Tmuxinator are related to their Ruby environments. By default, many Linux distributions and macOS come with a Ruby installed, but by superuser, requiring `sudo` to install and update gems. macOS (and likely many Linux flavors) ship with old and vulnerable Rubies.

In an attempt to rectify these issues, I propose that we evaluate the feasibility of shipping Tmuxinator **with** a recent Ruby by way of [Homebrew](https://brew.sh), [Snap](https://snapcraft.io), and [Flatpak](https://flatpak.org).

ethagnawl:
We could also consider porting the project to Crystal. 

Just kidding ... mostly. 😜

chris-fa:
Homebrew please. I only have Gem installed for this tool

nexeck:
Homebrew would be great :)

shivawu:
Homebrew please. Using gem means I cannot have `brew upgrade` to get the updates.

R4YM3:
bump

LandonSchropp:
I would really love to see this. It would be way more convenient to add `tmuxinator` to my `Brewfile`.

ethagnawl:
I'm not sure the best way to go about this, as I do not have a Mac and have never created a Homebrew package.

After a cursory search, it looks like this would require forking [homebrew-core](https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#push). Assuming that's correct ([other gems](https://github.com/sinclairtarget/homebrew-core) are doing this, so I'm pretty sure it is), one possible path forward would be for the tmuxinator organization to fork the homebrew-core repo and have whoever wants to take a pass at this submit a PR to that repository. I'm happy to do that, if you're interested in pursuing this, but we would need sign-off from one of the other maintainers, as I do not want to be on the hook for this if I'm not able to debug/test/release the changes myself.

In the meantime, a much simpler solution would be to use [brew-gem](https://github.com/soupmatt/brew-gem).

nexeck:
brew-gem does not work with Brewfile ;-)

nexeck:
I have created a formular for tmuxinator: https://github.com/nexeck/tap-tmuxinator

nexeck:
I will try to figure out, how to submit a new formula to the official repo, until that you could use my tap.

Any feedback is welcome :-)

[...]

Issue #529, opened 2017-05-22
Testing from within tmux fails


adamstrickland:
After merging https://github.com/tmuxinator/tmuxinator/pull/426, running tests from within a tmux session fails, due presumably to issues with [creating a new tmux session from within an existing tmux session](https://github.com/tmuxinator/tmuxinator/blob/master/spec/lib/tmuxinator/cli_spec.rb#L294).

rglassett:
What version of tmux are you using? Tests run fine for me under tmux 2.3 on macOS Sierra.

PR #682, opened 2019-02-05
Feature/Request Development environment

raiyuza:
While developing for tmuxinator i got increasingly frustrated with the
fact that i couldn't properly recreate bugs or environments.

So this is my attempt to fix it.

- Adds vagrant support for multi distro testing
- Adds docker support

This creates a way to develop locally more easily. Please read the
development section in CONTRIBUTING.md on how this all functions.

I also added to the gitignore file for some related and commonly used
files by developers.

This relates to
- #563
- #529

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ vendor/
 tags
 .env*
 .env
+Makefile
+.vagrant
+.ruby-version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,86 @@
     like OS version, gem versions, rbenv or rvm versions etc...
   * Even better, provide a failing test case for it.
 
+## Development
+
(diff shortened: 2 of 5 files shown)


raiyuza:
I might need some help here getting the syntax correct, i am only getting so far with rubocop. Why does it parse my bash in HEREDOC?

ethagnawl:
Have you tried disabling RuboCop within the HEREDOC?

Something like:

```
rbenv_script = <<~SCRIPT
# rubocop:disable all
echo "do a thing"
# rubocop:enable all
SCRIPT
```

Additionally, perhaps the Ruby version should be pulled from an ENV var and fall back to the minimally supported version if it's not present? This would prevent people from having to remember to checkout the affected files after switching Ruby versions.

ethagnawl:
There hasn't been any activity here in some time. If you're interested in revisiting this, feel free to reopen.

PR #692, opened 2019-04-26
add tmux 2.9 support

teddyhwang:
tmuxinator works fine with tmux 2.9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+### Misc
+- add support for tmux 2.9
+
 ## 0.16.0
 ### Bugfixes
 - fix wemux class_eval error (#590)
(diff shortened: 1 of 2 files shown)


Croissong:
I ran into the same problem, thanks for the quick PR :)
On a side note: Is it possible to configure `tmuxinator start` to ignore the supported versions?
I'm running it in my xinit script and the "unsupported version" prompt caused it to crash silently.

teddyhwang:
I think you can just add the `--suppress-tmux-version-warning` flag

ethagnawl:
Thanks, @teddyhwang!

Would you mind adding an entry to CHANGELOG.md per [the contributing guidelines](https://github.com/tmuxinator/tmuxinator/blob/master/CONTRIBUTING.md#pull-requests)?

teddyhwang:
@ethagnawl oof, my bad. Added!

PR #694, opened 2019-04-26
1.0.0

ethagnawl:
- add support for tmux 2.9


ethagnawl:
Thoughts, @tmuxinator/tmuxinator? If the bump to 1.0.0 seems excessive or rushed, we can make this 0.17.0 and talk about a better time to cut 1.0.0.

Issue #691, opened 2019-04-22
Procfile parsing


jgandt:
**Is your feature request related to a problem? Please describe.**
I would like to use tmuxinator to launch processes listed in a Procfile.

**Describe the solution you'd like**
As an extension of tmuxinator, I would like to be able to launch processes listed in a Procfile. I would like to launch each process in a named tmux window or pane. It seems a nice extension to tmuxinator based on an industry defined spec (the Procfile).

**Describe alternatives you've considered**
There are theoretically some tools out there to accomplish this, but none have the polish nor pedigree of tmuxinator. They also don't have the same amount of configuration available.
  
I could write a separate utility myself, but it seems to fit in tmuxinator's wheelhouse. In addition it would get more visibility if it was a feature in such a widely used product.

I could convert my Procfile to a tmuxinator config, but then I have to keep the tmuxinator config up to date. A conversion to strictly using tmuxinator is not an option as the Procfile is used in production by Heroku. Also, other members of my team may not leverage tmuxinator. They might instead use foreman or overmind.

**Additional context**
I'm happy to take a crack at writing the extension, but I figured I should ask if it's something you would consider valuable before I put time into it. If I do write it, I would prefer to have your input on the design.

ethagnawl:
Hi, @jgandt. This is an interesting idea and, frankly, I'm surprised variations of this (JSON, INI, S-EXP, etc.) aren't requested more frequently.

I'll think this over and welcome any thoughts you or others have about 1) whether or not this is worth considering and 2) how we might go about implementing it.

To quickly follow up on #2, I could see the introduction of a plugin based architecture which would require a new abstract base class sitting somewhere between or replacing the existing Config/Project/others? classes which could be extended to introduce new config file formats. From there, and depending on how people feel about this and the potential maintenance burden, we could have a select few "official" plugins provided by tmuxinator and allow users to use third-party plugins (which could eventually be rolled into/officially supported by tmuxinator).

jgandt:
That sounds great! Thanks for even considering it!

If you don't have any objections, I'll see if I can put some code together to do as you suggest.

I don't have any experience in the codebase. If I hit a wall I will wait to hear your decision before I ask any further advice on code architecture.

Thanks again!

jgandt:
Ok I got some time and I took a very quick shot at the core mechanism.
#697 

This is 100% WIP but I figured we could have a conversation about this before I get too much further.

What are your thoughts? Am I on the right track or would you like to see things done differently?

PR #697, opened 2019-04-30
WIP - Alternate config formats (Do Not Merge) See: #691

jgandt:
This is just a Q&D proof of concept to show the parsing mechanism.

I have not written tests. I have not squashed commits. Lots of other stuff was left out so that we can initiate a conversation about the mechanism and architecture.


lib/tmuxinator/project.rb
```diff
@@ -41,33 +41,17 @@ class Project
...
-        args = options[:args] || []
-        @settings = parse_settings(args)
-        @args = args
-
-        content = Erubis::Eruby.new(raw_content).result(binding)
-        YAML.safe_load(content, [], [], true)
+        config_file_type = options[:type] || 'default'
+        # now must pass in type
+        # I temporarily hard coded it for getting 'procfile' type up and running
+        yml = Parser.new(path, options, 'procfile').build_yaml
```
jgandt: Obviously the 'procfile' type is temporarily hard coded. I didn't want to dive into cli changes until we get the parsing mechanism close to what we want.

jgandt:
Also this is not 100% working for the Procfile, so I still have some work to do there. I'm not entirely sure what I did wrong, but the yaml I created in `Parsers::Procfile` doesn't quite do the trick.

ethagnawl:
Hey, @jgandt. 

This is a great start and I'm definitely interested in seeing this fleshed out.

A few quick thoughts:
- As this stands, it looks this approach will coerce any/all config files _into_ YAML. Right? This makes sense as a starting point and/or bridge to the final implementation, but my initial thought was that we'd have a config file class/hash/similar which provides the uniform interface to the config data. Granted, that would probably complicate the implementation of new config file types, but would prevent tmuxinator from arbitrarily leaning on YAML. I could be swayed, though, so I'd definitely be interested in your thoughts. (One complication with my approach is with how ERB interpolation would be handled -- I'll need to think some more about how the parsers would handle that and maybe look at how projects like Sprockets handle that (e.g. some-library.js.erb, which first interpolates any Ruby code contained in the JS file).)
- Following up on that, I think YAML should lose its special status and be treated as Just Another Parser Class (:stuck_out_tongue_winking_eye:). So, `Default` would become `YAML`.

Any thoughts on the above, @tmuxinator/tmuxinator?

jgandt:
So I'm parsing it into "yaml." It's really just a ruby representation of the current config file format. That is, YAML can be directly translated into ruby objects. Therefore the current config file format already provides the uniform interface to the config data. So really I'm just converting the Procfile into that pre-defined interface. I am happy to wrap a class or code around it so we can provide some better documentation for that format. That being said, I believe you already have some documentation around valid configuration files in the form of examples. However if we still want to codify it in the code, I'm happy to so.

It just so happens that the Procfile format is parseable by YAML parsers. The format is:
```
process_name_1: process command 1
process_name_2: process command 2
...
```
So I'm really just using a quirk of the format to conveniently parse the file. Then I stuff it into the "yaml" variable. I'll rename some variables and methods and we can see if that makes things clearer.

jgandt:
> IMO, this would be great. It'd be nice to have a defined API and would make it simple for others to introduce new parsers.

You got it! I'll try to put it together tonight.

jgandt:
Ok I think I'm going to do some refactoring in the project. I'd like to reclaim the `Config` module to handle the actual configuration object that will come out of parsed files.

Right now, the `Config` module is really more of a config file finder. I will try to slowly refactor the code to reclaim the module if that is ok.

If you would prefer to leave it as is, I can find another name for the config object. `ProjectConfig` or something would work. It might even be more clear on the intent.

I took a small first shot at some refactoring that I think is a good one even if we don't rename the module. see: #700.

adamstrickland:
> It'd be nice to have a defined API and would make it simple for others to introduce new parsers. 

I think that this is a possible long-term goal.  However, maybe before we start down that road, we wait to see what kind of demand we're looking at.  I have created https://github.com/tmuxinator/tmuxinator/issues/703 to solicit community feedback on a more general solution.

I think I would prefer incorporating *just* the Procfile parsing with minimal refactoring as the right step at this time.

PR #700, opened 2019-05-03
Move Config.version to TmuxVersion.version

jgandt:
This method better fits in TmuxVersion where it's purpose matches the
module's purpose.

### DIFF
--- a/spec/lib/tmuxinator/cli_spec.rb
+++ b/spec/lib/tmuxinator/cli_spec.rb
@@ -252,8 +252,8 @@ describe Tmuxinator::Cli do
   describe "#start" do
     before do
       ARGV.replace(["start", "foo"])
+      stub_tmux_version(1.9)
       allow(Tmuxinator::Config).to receive_messages(validate: project)
-      allow(Tmuxinator::Config).to receive_messages(version: 1.9)
       allow(Kernel).to receive(:exec)
     end
 
@@ -304,8 +304,8 @@ describe Tmuxinator::Cli do
   describe "#stop" do
     before do
       ARGV.replace(["stop", "foo"])
+      stub_tmux_version(1.9)
       allow(Tmuxinator::Config).to receive_messages(validate: project)
-      allow(Tmuxinator::Config).to receive_messages(version: 1.9)
       allow(Kernel).to receive(:exec)
     end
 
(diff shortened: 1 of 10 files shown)


jgandt:
This is a small refactor. I am looking to do some other refactoring of the current Config class so this was a nice isolated refactor that I wanted to get done.

Thoughts?

adamstrickland:
It would be helpful if you could provide some context on why you're doing this refactor.  I imagine it has something to do with your implementation of https://github.com/tmuxinator/tmuxinator/pull/697, but it's not clear how this furthers that objective.  As I noted in [#697](https://github.com/tmuxinator/tmuxinator/pull/697#issuecomment-489092610), I think that keeping the scope to a minimum would be the quickest way to get that feature into the project.

That said, I'm a little uncomfortable refactoring the `Config` in place.  In https://github.com/tmuxinator/tmuxinator/pull/697 you mention using the name `ProjectConfig`; again, it's not clear to me what this is supposed to hold (and therefore what it should be named), but I would lean towards building out a separate construct and then we'll worry about merging in later if that makes sense.

spec/lib/tmuxinator/cli_spec.rb
```diff
@@ -252,8 +252,8 @@
   describe "#start" do
     before do
       ARGV.replace(["start", "foo"])
+      stub_tmux_version(1.9)
```
adamstrickland: I think I would prefer to leave the stub in-place in the spec, rather than externalize it to a separate file, since we're not looking at having to stub out more than the one method or doing this in a bunch of spec suites.  Perhaps using `let`s would make this more DRY, fi that's what you're going for, e.g.

```ruby
describe "#stop" do
  before do
    allow(Tmuxinator::Config).to receive_messages(version: version)
  end

  context "with tmux 1.9" do
    let(:version) { 1.9 }

    it { is_expected.to be_truthy }
  end
end
```

jgandt:
That's totally fine. I'll close this for now.

I saw an opportunity for some small refactors and thought I would go for them. I guess it just seemed awkward to have the version, asset path, and project loading stuff all in the same class/module. I think this is really only a code clarity and organization refactor. It is certainly not necessary.

Please note, this is true for #700, #701, #702.

I will close them all for now and just get on with the feature development.

PR #701, opened 2019-05-03
Isolate asset path methods

jgandt:
This moves asset path methods out of Config and into their own AssetPath class.

This is dependent upon #700. It was branched off of that branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Misc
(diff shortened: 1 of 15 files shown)


jgandt:
This is another small refactor that I think stands alone very well. Let me know what you think!

jgandt:
This is based off of #700 so the conflicts are due to that. They should resolve once #700 is merged.

jgandt:
See: https://github.com/tmuxinator/tmuxinator/pull/700#issuecomment-489096929

PR #702, opened 2019-05-03
Move project load out of config validate

jgandt:
This separates the concerns of locating and parsing the config file.

This will allow us to better isolate the Config class as a whole.

Config.validate now returns a string to the location of the config file.

Consumers of Config.validate are now responsible for executing their own
Project.load.

This was branched from #701 and is slightly dependent thereof. It can easily be extricated from that dependency if needed.

Any conflicts should resolve once #701 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Misc
+Small refactor to move Config.version to TmuxVersion.version
+
+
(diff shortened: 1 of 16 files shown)


jgandt:
Let me know what you think!

jgandt:
See: https://github.com/tmuxinator/tmuxinator/pull/700#issuecomment-489096929

Issue #705, opened 2019-05-15
tmuxinator.zsh:21: command not found: compdef


HaleTom:
**Describe the bug**
When starting my shell, I see:

    tmuxinator.zsh:21: command not found: compdef

**To Reproduce**
```
% zsh -f
svelte% source /home/ravi/.local/share/gem/gems/tmuxinator-1.1.0/completion/tmuxinator.zsh
/home/ravi/.local/share/gem/gems/tmuxinator-1.1.0/completion/tmuxinator.zsh:21: command not found: compdef
```
**Expected behavior**

No output

**Environment**
 - tmuxinator Version: tmuxinator 1.1.0
- tmux Version: tmux 2.9
 - Ruby Version: ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]
 - OS: Linux svelte 5.0.9-2-MANJARO #1 SMP PREEMPT Sun Apr 21 07:11:08 UTC 2019 x86_64 GNU/Linux

**Fix:**
```
svelte% autoload -Uz compinit; compinit
svelte% source /home/ravi/.local/share/gem/gems/tmuxinator-1.1.0/completion/tmuxinator.zsh
svelte%
```

PR #706, opened 2019-05-15
Load compinit before compdef

HaleTom:
Fix command not found: compdef #705

### DIFF
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -18,6 +18,7 @@ _tmuxinator() {
   return
 }
 
+autoload -Uz compinit; compinit
 compdef _tmuxinator tmuxinator mux
 alias mux="tmuxinator"
 


ethagnawl:
This looks good to me, but I don't use zsh.

Does anyone @tmuxinator/tmuxinator who's familiar with zsh mind having a look?

rglassett:
I'm only a casual zsh user, so take with a grain of salt; but as far as I can tell this isn't the conventional way to define completion functions. `compinit` is meant to be called from the user's `~/.zshrc`, and the definitions are loaded from files in the `$fpath` variable. [Git completion](https://github.com/git/git/blob/master/contrib/completion/git-completion.zsh) looks to be a good example.

I'll need to read up on it more, but I think the actual completion file should look something like this:

```
# tmuxinator/completion/tmuxinator.zsh

_tmuxinator() {
  # logic
}

_tmuxinator
```

To install, users would copy the file to `$ZDOTDIR/functions/Completion/_tmuxinator` and add to their `~/.zshrc`:

```
fpath=($ZDOTDIR/functions/Completion $fpath)
autoload -Uz compinit
compinit

# optional alias
alias mux="tmuxinator"
compdef mux=tmuxinator
```

I'll try to find time over the weekend to read up and do a refactor.

Rahlir:
I believe this is the correct approach:
https://github.com/zsh-users/zsh-completions/blob/master/src/_tmuxinator

Rahlir:
Note that for such completions, the file should be placed in a directory (typically used for zsh completions in general) that is present in your `$fpath`

ethagnawl:
@HaleTom @rglassett What is everyone thinking about the existing vs. proposed solutions?

doronbehar:
@ethagnawl @Rahlir is correct, I was about to file an issue / PR about this as well: In short, ZSH completions aren't files you source in your `~/.zshrc`. You are supposed to put them in your `$fpath` and they'll be loaded on demand. The better PR should be this:

```patch
diff --git c/completion/tmuxinator.zsh i/completion/tmuxinator.zsh
index e77f1ed..e291eac 100644
--- c/completion/tmuxinator.zsh
+++ i/completion/tmuxinator.zsh
@@ -1,3 +1,4 @@
+#compdef _tmuxinator tmuxinator
 _tmuxinator() {
   local commands projects
   commands=(${(f)"$(tmuxinator commands zsh)"})
@@ -18,8 +19,6 @@ _tmuxinator() {
   return
 }
 
-compdef _tmuxinator tmuxinator mux
-alias mux="tmuxinator"
 
 # Local Variables:
 # mode: Shell-Script
```

Along with a README update as well.

doronbehar:
BTW AFAIK fish and bash are the same - you don't need to source anything, just put it in a location the shell looks for completions in it. Usually that's taken care of by the package.

ethagnawl:
Thanks for this, @doronbehar.

> The better PR should be this:

Does anyone have thoughts on this? It'd be nice to get this sorted out and ship it in the next release.

ethagnawl:
@doronbehar @HaleTom @rglassett Any thoughts on the above? I'm hoping to cut a release today or tomorrow and it'd be nice to get this in for the zsh folks.

doronbehar:
I've stated my opinion in my previous reply - https://github.com/tmuxinator/tmuxinator/pull/706#issuecomment-546596771

patrick96:
I didn't even know tmuxinator had its own zsh completion, I always used the `zsh-completions` package.

But @doronbehar's changes look good to me and the zsh-completions project's [guide](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command) seems to agree.

But the completions file should be named `_tmuxinator` and the README needs to be adjusted to tell the user to put it into a directory in the `fpath`.

ethagnawl:
Please feel free to update this PR or submit a new one.

PR #717, opened 2019-08-10
add zshell completions for command aliases

remyoudemans:
Up until now, in zsh you had to type out the full command
(like "tmuxinator start") for tab completions to work.
This makes it work for aliases too (like "tmuxinator s").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+### Misc
+- add zshell completions for command aliases
+
 ## 1.1.1
 ### Bugfixes
 - increase min XDG version in gemspec in order to exclude broken release (#708)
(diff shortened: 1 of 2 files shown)


ethagnawl:
Would anyone who uses zsh mind reviewing this? @tmuxinator/tmuxinator

magicalbanana:
@remyoudemans can you elaborate? Do you mean, when you did `tmuxinator ...` it didn't display the autocompletions?

remyoudemans:
@magicalbanana no. I made the project completions work for command aliases.
What already existed was that if I typed `tmuxinator start ` and then pressed tab, it would complete to one of my projects like `tmuxinator start myProject`.

However, if I used the alias for `start` which is `s` and then pressed tab (`tmuxinator s ` and then tab), I would get no completions.

All I did was make it check for completions not only when a full command is written (`start`, `edit` ...) but also when you use aliases (`s`, `e` .....)

magicalbanana:
@remyoudemans gotcha. Thank you for the clarification. This looks good to me.

PR #720, opened 2019-09-25
Specify socket in 'tmux_has_session?'

### LINKED ISSUE #719: Tmuxinator repeats the initialization commands on attach
**Describe the bug**
Tmuxinator correctly sends keys to a tmux window upon creation of the session. However, subsequent calls of `tmuxinator SESSION` not only attaches to the running session but also resends the keys. This seems to happen only when `tmux_options` are included.

**To Reproduce**
Use this configuration:

```yaml
# ~/.tmuxinator/test.yml

name: test
root: ~/

tmux_options: -L test

windows:
  - test_window:
      - echo $(date)
```
then type
```bash
tmuxinator test
```
detach from session and type again
```bash
tmuxinator test
```

**Expected behavior**
The second call of `tmuxinator test` should not send `echo $(date)`.

**Environment**
 - tmuxinator Version: 1.1.1
 - tmux Version: 2.9a
 - Ruby Version: 2.6.4p104
 - OS: Archlinux

urbainvaes:
This fixes #719. The socket must be specified via 'socket_name' and not
via '

I tested  with the following configuration file,

```yaml
name: test
root: ~/

socket_name: foo

windows:
  - test_window:
      - echo $(date)
```
and it worked as expected.

### DIFF
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -170,7 +170,7 @@ module Tmuxinator
       # to server: Connection refused" error message and non-zero exit status
       # if no tmux sessions exist.
       # Please see issues #402 and #414.
-      sessions = `#{tmux_command} ls 2> /dev/null`
+      sessions = `#{tmux_command}#{socket} ls 2> /dev/null`
 
       # Remove any escape sequences added by `shellescape` in Project#name.
       # Escapes can result in: "ArgumentError: invalid multibyte character"


ethagnawl:
Thanks for this, @urbainvaes. 

We'll need to fix the broken tests (the quick solution is just adding the socket flag/name to the expected cmd string) and update the CHANGELOG before merging this. (I'm happy to help.)

However, now that I'm looking at the Project class, I'm wondering why we're not using `#tmux` instead of `#tmux_command` in `tmux_has_session?`, which already includes `#socket`? (We should also make the name of the socket in sample.yml something more meaningful than `foo`, like `a-custom-socket`.) `#tmux` adds additional options to the call to `tmux` but I don't _think_ this will break anything -- aside from the tests which aren't expecting the extra arguments. If you want to investigate this, feel free, otherwise I can take a pass at it.

urbainvaes:
Thanks @ethagnawl for the quick response, and sorry for the broken tests. I haven't looked at the code very carefully yet, but it seems that using `#tmux` instead of `#tmux_command` is a good idea.

Is there anything specific I can do? If you let me know how to run the tests (I haven't used ruby before), I can take care of fixing the ones that broke.

ethagnawl:
Closing in favor of #723.

PR #723, opened 2019-09-26
720 redux

### LINKED ISSUE #719: Tmuxinator repeats the initialization commands on attach

ethagnawl:
Fixes #719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## UNRELEASED
+- prevent commands from being re-run when re-attaching to session using custom
+socket (#719)
+
 ## 1.1.1
 ### Bugfixes
 - increase min XDG version in gemspec in order to exclude broken release (#708)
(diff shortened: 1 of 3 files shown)


ethagnawl:
Would anyone mind reviewing these changes? (/CC @tmuxinator/tmuxinator) I can't imagine using the `#tmux` method (and its additional flags), instead of the more basic `#tmux_command`, will introduce unwanted behavior but it's possible I'm missing something.

rglassett:
Come to think of it, is there a compelling reason to have a `tmux_options` config at all? Why shouldn't that get merged with `tmux_command`? e.g. `tmux_command: tmux -L foo`

I can't think of a reason why it'd be useful to declare them separately.

ethagnawl:
Thanks for reviewing this, @rglassett!

I agree with the changes you're proposing. I tried a very basic pass at them which broke some tests and, seeing as I don't have time to make a proper pass at it right now, I'm going to punt. I'll create an issue and feel free to take a pass, if you'd like to.

Otherwise, I'm going to merge this and cut a release, as this fix should really get out to end users ASAP.

PR #729, opened 2019-10-06
Correct edge tmux version detection

### LINKED ISSUE #728: Incorrect version detection
Running tmuxinator with Tmux built from`HEAD` is problematic for non-trivial layouts.
Upon launch an error about unsupported option `default-path` is output by Tmux server.

A `HEAD` Tmux currently reports itself as `next-3.1` (`$ tmux -V`), which fails https://github.com/tmuxinator/tmuxinator/blob/6442103d03c3c95e73e86db2044cf35d7a13f55b/lib/tmuxinator/config.rb#L63-L73 logic

```ruby
>> 'next-3.1'.split(" ")[1]
=> nil
```

Hence, later on in tmuxinator where checks are made against `version < 1.8` it incorrectly assumes that I'm running ye'olde Tmux.

aleksandrs-ledovskis:
Closes #728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Bugfixes
+- correct edge tmux version detection (#728)
+
 ## 1.1.2
 ### Bugfixes
 - prevent commands from being re-run when re-attaching to session using custom
--- a/lib/tmuxinator/config.rb
+++ b/lib/tmuxinator/config.rb
@@ -67,7 +67,7 @@ module Tmuxinator
           if tmux_version == "master"
             TMUX_MASTER_VERSION
           else
-            tmux_version.to_f
+            tmux_version.to_s[/\d+(?:\.\d+)?/, 0].to_f
           end
         end
       end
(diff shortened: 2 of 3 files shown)


ethagnawl:
Thanks for putting this together, @aleksandrs-ledovskis. This is an ideal PR. :)

However, this approach will break for versions which include post-fixed text. 

For example:


```
  1) Tmuxinator::Config#version text postfixed should eq 2.9
     Failure/Error: it { is_expected.to eq 2.9 }
     
       expected: 2.9
            got: 0.0
     
       (compared using ==)
     # ./spec/lib/tmuxinator/config_spec.rb:174:in `block (4 levels) in <top (required)>'

Finished in 1.32 seconds (files took 0.49474 seconds to load)
253 examples, 1 failure
```

This all points to a larger problem, though and I've opened an [issue](#730) which covers that.

In terms of getting this issue addressed, though, maybe you could make the regex more permissive and add an additional test case that asserts the existing behavior (which, regrettably, coerces "2.9a" to 2.9) holds?

aleksandrs-ledovskis:
@ethagnawl I have addressed your findings in new version of PR. Please check

PR #737, opened 2019-12-04
Fix completion related issues

doronbehar:
@ethagnawl if it wasn't noticeable, my comment https://github.com/tmuxinator/tmuxinator/pull/706#issuecomment-546596771 had a patch that would close #706 and #705. But this PR includes an update to the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,27 +51,24 @@ earlier versions. Refer to the FAQ for any odd behaviour.
 
 ## Completion
 
-Download the appropriate completion file from the repo and `source` the file.
-The following are example where the completion file has been downloaded into
-`~/.bin`.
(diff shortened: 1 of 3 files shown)


HaleTom:
Thanks a bunch for this!  My shell now loads without an error message, and I'm a very happy chappy 😄

ethagnawl:
@doronbehar Any thoughts on @HaleTom's comments? 

Also, could you remind me why the alias was removed in this PR? I thought that was covered either here or in the related issue, but I don't see mention of the rationale. Either way, users are expecting it to be provided based on the README, so if for some reason it's not possible to have the completion file define it, we should update the README to note that.

PR #742, opened 2019-12-26
Upgrade Thor runtime dependency

### LINKED ISSUE #739: Crash with ruby-thor 1.0.1
```
Traceback (most recent call last):
        8: from /usr/bin/tmuxinator:23:in `<main>'
        7: from /usr/lib/ruby/2.6.0/rubygems.rb:304:in `activate_bin_path'
        6: from /usr/lib/ruby/2.6.0/rubygems.rb:304:in `synchronize'
        5: from /usr/lib/ruby/2.6.0/rubygems.rb:305:in `block in activate_bin_path'
        4: from /usr/lib/ruby/2.6.0/rubygems/specification.rb:1417:in `activate'
        3: from /usr/lib/ruby/2.6.0/rubygems/specification.rb:1435:in `activate_dependencies'
        2: from /usr/lib/ruby/2.6.0/rubygems/specification.rb:1435:in `each'
        1: from /usr/lib/ruby/2.6.0/rubygems/specification.rb:1446:in `block in activate_dependencies'
/usr/lib/ruby/2.6.0/rubygems/dependency.rb:313:in `to_specs': Could not find 'thor' (>= 0.15.0, ~> 0.19) - did find: [thor-1.0.1] (Gem::MissingSpecVersionError)
Checked in 'GEM_PATH=/home/nycko/.gem/ruby/2.6.0:/usr/lib/ruby/gems/2.6.0', execute `gem env` for more information
```

ethagnawl:
Upgrade to Thor ~> 1.0.0 in order to support Arch package.

Closes #739

### DIFF
--- a/tmuxinator.gemspec
+++ b/tmuxinator.gemspec
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.4.6"
 
   s.add_dependency "erubis", "~> 2.6"
-  s.add_dependency "thor", "~> 0.19", ">= 0.15.0"
+  s.add_dependency "thor", "~> 0.19", ">= 0.15.0", "< 1.0.0"
   s.add_dependency "xdg", "~> 2.2", ">= 2.2.5"
 
   s.add_development_dependency "activesupport", "< 5.0.0" # Please see issue #432
--- a/tmuxinator.gemspec
+++ b/tmuxinator.gemspec
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.4.6"
 
   s.add_dependency "erubis", "~> 2.6"
-  s.add_dependency "thor", "~> 0.19", ">= 0.15.0", "< 1.0.0"
+  s.add_dependency "thor", "~> 1.0"
   s.add_dependency "xdg", "~> 2.2", ">= 2.2.5"
 
   s.add_development_dependency "activesupport", "< 5.0.0" # Please see issue #432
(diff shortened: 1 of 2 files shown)


ethagnawl:
Thoughts @tmuxinator/tmuxinator?

tmuxinator.gemspec
```diff
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.4.6"
 
   s.add_dependency "erubis", "~> 2.6"
-  s.add_dependency "thor", "~> 0.19", ">= 0.15.0", "< 1.0.0"
+  s.add_dependency "thor", "~> 1.0"
```
adamstrickland: what changed that this is the right thing to do?  perhaps we should document whatever it is in the changelog?

ethagnawl: The initial "fix" didn't actually address the underlying "problem". Thor releasing v1+ breaks tmuxinator on Arch because the package's build description file doesn't specify any version constraints for the Arch ruby-thor package -- it uses Arch packages in place of tmuxinator's runtime dependencies as managed via RubyGems.

In this case, the actual fix which accommodates the Arch package is pretty minimal and tmuxinator was already compatible with Thor's new API, but it is debatable whether this sort of issue should be addressed in tmuxinator or the Arch package.

ethagnawl: Any additional thoughts, @adamstrickland? The Arch folks seem happy with this and it should be a seamless upgrade for people using RubyGems.

PR #744, opened 2020-01-21
Restore alias mux for zsh

mbbo:
🚩 **Summary**
In this specific commit : https://github.com/tmuxinator/tmuxinator/pull/737/commits/f149defef09251a9e2177626ca888e3041f5a11c#diff-10ce7fe3ff7170249062abe941ccd519L22

Aliase for mux in zsh has been removed, so I can't use the mux command in my terminal

⚠️ **Actual Result**
zsh_mux

✔️ **Expected Result**
Can use the mux command

### DIFF
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -20,6 +20,8 @@ _tmuxinator() {
   return
 }
 
+complete -F _tmuxinator tmuxinator mux
+alias mux="tmuxinator"
 
 # Local Variables:
 # mode: Shell-Script


ethagnawl:
Hi, @mbbo. That alias was removed recently by #737. I've asked the author to remind me why that was the case, so we'll have to wait and see what they say -- I don't use ZSH, so I can't offer much help here. See my comment in that issue, though. If for some reason it's not possible to set the alias in the ZSH completion file, the README should be updated to note that.

In the meantime, can you add the alias to your .bashrc (or similar)?

mbbo:
Thanks you for your answer. I do have a workaround possbile by adding the aliase in the .zshrc, it works. I wait also the reason of why it was deleted

doronbehar:
Responding here to @HaleTom 's comment (https://github.com/tmuxinator/tmuxinator/pull/737#issuecomment-574179372) as here more people are involved:

@HaleTom I'm not sure what's wrong with your setup. The only thing I can ask you to make sure is to check if you can find the file `_tmuxinator` in your `$fpath` - this is the responsibility of the package maintainer, I doubt that `gem` takes care of this..

@mbbo The reason it was deleted is because the completion file is deleted on demand and it's usually not common for packages to force such aliases on users. I'd suggest the maintainers to put a note regarding this change in the README.

Moreover, if you just put the alias there in the completion file and it's installed correctly (NOTE @HaleTom) and I mean by that, that it's installed as `_tmuxinator` to a directory in your `$fpath` (and not just `source`d in your `.zshrc`!), then, the alias wouldn't work because the completion file is loaded only on demand - only until you type `tmuxinator <tab>`.

Please trust me on this guys, I'm just trying to help. I'm not sure how familiar you are with command line programs and packaging standards but I'm telling you from experience this is how this works in most projects. I know it's not what you are used to with tmuxinator (which it's completion is critical for it's UX) but this is what's **right**.

mbbo:
There is nothing to do anymore on this pull request so i close it

PR #748, opened 2020-03-27
add shorthand alias for zsh completion

yhchen0906:
The [shorthand alias](https://github.com/tmuxinator/tmuxinator#shorthand) for zsh completion is missing.

### DIFF
--- a/completion/tmuxinator.zsh
+++ b/completion/tmuxinator.zsh
@@ -20,6 +20,7 @@ _tmuxinator() {
   return
 }
 
+alias mux="tmuxinator"
 
 # Local Variables:
 # mode: Shell-Script


ethagnawl:
Hi, @yhchen0906.

Please see the conversation in #744 about this feature. The consensus is that the alias should be added to the user's .zshrc or similar.

The README is currently outdated based on the decision made in that PR and the linked PRs and issues, so I will update it.

yhchen0906:
Hi, @ethagnawl 
Thanks for your reply!

Let me make sure I get it right. According to the conversation, the consensus is that alias should not appear in the completion definition, which means the shorthand alias in bash and fish should also be removed.

Please let me know if I misunderstood anything, thank you.

ethagnawl:
That's correct. See #749.

I'm going to close this for now, but let me know if there's anything else you think is worth discussing.

PR #779, opened 2020-06-23
Add support for tmux 3.2

ethagnawl:
tmux 3.2 is still an RC, but this is in preparation for its release.


alk-mandrianarijaona:
it would be great to have this

PR #793, opened 2020-12-06
Proof of Concept for Erubi

jimtahu:
Just realized your offer to skim the proof of concept mentioned in tmuxinator#792 needed a matching pull request

### DIFF
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -169,7 +169,7 @@ module Tmuxinator
       def generate_project_file(name, path)
         template = Tmuxinator::Config.default? ? :default : :sample
         content = File.read(Tmuxinator::Config.send(template.to_sym))
-        erb = Erubis::Eruby.new(content).result(binding)
+        erb = eval(Erubi::Engine.new(content).src, binding)
         File.open(path, "w") { |f| f.write(erb) }
         path
       end
--- a/lib/tmuxinator/cli.rb
+++ b/lib/tmuxinator/cli.rb
@@ -169,7 +169,8 @@ module Tmuxinator
       def generate_project_file(name, path)
         template = Tmuxinator::Config.default? ? :default : :sample
         content = File.read(Tmuxinator::Config.send(template.to_sym))
-        erb = binding.instance_eval(Erubi::Engine.new(content).src)
+        erubi_content = Erubi::Engine.new(content).src
+        erb = binding.instance_eval(erubi_content)
         File.open(path, "w") { |f| f.write(erb) }
         path
       end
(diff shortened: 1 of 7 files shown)


ethagnawl:
@jimtahu I didn't mean to encroach with those commits, but I wanted to bring this PR up-to-date and re-trigger the build to see what Code Climate was complaining about.

ethagnawl:
I'm inclined to ignore these Code Climate failures and am considering increasing the default complexity threshold which is causing them.

jimtahu:
> @jimtahu I didn't mean to encroach with those commits, but I wanted to bring this PR up-to-date and re-trigger the build to see what Code Climate was complaining about.

No complaints from me, thanks for the pull up

lib/tmuxinator/cli.rb
```diff
@@ -169,7 +169,8 @@ def config_path(name, local = false)
       def generate_project_file(name, path)
         template = Tmuxinator::Config.default? ? :default : :sample
         content = File.read(Tmuxinator::Config.send(template.to_sym))
-        erb = Erubis::Eruby.new(content).result(binding)
+        erubi_content = Erubi::Engine.new(content).src
+        erb = binding.instance_eval(erubi_content)
```
akofink, line 173: What does using `instance_eval` instead of `eval` do? `binding` is a new instance every time it's called.

akofink (approved):
Works well! 🥇 One small comment/question inline.

lib/tmuxinator/cli.rb
jimtahu: > What does using `instance_eval` instead of `eval` do? `binding` is a new instance every time it's called.

It's been so long since I last looked at this that I'm afraid the best I can remember was something about limiting scope and being able to dump the object for debugging.

akofink: It makes sense to me that `instance_eval` is safer/more limited scope.

I notice we're not consistent with `eval` vs. `instance_eval` in this PR, [like in project.rb](https://github.com/tmuxinator/tmuxinator/pull/793/files#diff-e013992ae474239aef1e409cb960c112e5c981d8c6fedacb5451245b6fbfa289R107). Maybe we should be using `instance_eval` there as well?

Maybe unrelated, but [this SO answer](https://stackoverflow.com/a/46567196) shows how `instance_eval` can be used with a context object and instance variables (i.e. `<%= @foo %>`) in the template, which is common in Rails, for example. I don't think we're doing any of that in this repo. 🤔

akofink:
I'm going to merge this PR soon for all the reasons outlined in the linked issue. It appears to work great and passes all our testing after I recently rebased it with master. Let's finally merge it!

PR #820, opened 2021-05-18
fix error when parsing config file

jschank:
YAML parser returns wrong number of arguments(given 4, expected 1),
this fix removes the extra parameters, while still passing in the
allowAliases flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## Unreleased
+- Fix config file parsing error: wrong number of arguments (given 4, expected 1) (#819)
+
 ## 2.0.3
 ### Misc
 - Add Ruby 3 to the Travis test matrix
--- a/lib/tmuxinator/project.rb
+++ b/lib/tmuxinator/project.rb
@@ -48,7 +48,7 @@ module Tmuxinator
         @args = args
 
         content = Erubis::Eruby.new(raw_content).result(binding)
-        YAML.safe_load(content, [], [], true)
+        YAML.safe_load(content, aliases: true)
       rescue SyntaxError, StandardError => error
         raise "Failed to parse config file: #{error.message}"
       end


ethagnawl:
Thanks for putting this together @jschank. Do you or anyone else know what the underlying problem is? I've only been able to dedicate a few minutes to this, but haven't been able to reproduce it and don't understand why the CI passes. In the meantime, until someone figures out what the underlying problem is, I wonder if `YAML.method(:safe_load).arity` could be used to paper over the problem?

jschank:
I think the problem is that the YAML lib maybe changed its signature.
It is what throws the incorrect number of parameters exception.

In the tmuxinator code, it looks like the first and last parameters are significant, the middle two were just empty arrays.

It could be that the YAML lib has a better alternative method to call. But this fix seems to work ok.

John

ethagnawl:
There's been [a lot of churn in the Psych library](https://github.com/ruby/psych/releases) relating to this method, but I'm still not clear on what exactly changed (in libyaml or Psych) and when.

> But this fix seems to work ok.

Right, but it breaks the CI build. Granted, those failures were for versions of Ruby which have been EOL'd. If that winds up being the "best" fix, then we'll need to properly drop support for those versions of Ruby.

jschank:
I suppose the call could be wrapped, the cause could be my attempt to use named parameters. That might have been added in a relatively recent ruby version. So old versions might be able to use one call, and newer ones, a different call.

Sadly, this gets beyond my depth with ruby though.

ethagnawl:
Any thoughts on this, @tmuxinator/tmuxinator? I haven't had time to do a deep dive on the underlying issue or whether we should integrate this PR as-is and deprecate versions of Ruby which fail CI as a result or try to find a more backwards compatible approach.

jschank:
Hmm, my opinion is this:
1. I think we should try to be backward compatible. So I don’t think this PR is actually ready to be merged yet.
2. I think the fault (so to speak) is with the Psych library. They seem to have introduced a backward incompatible change.
3. IF there are desirable features in the new Psych release we should try to integrate with it if possible
4. IF there is no compelling reason to use the new release - Can the bundle peg the Psych release to the previous (functioning version)? Admittedly, I don’t know if ruby dependencies can be pegged like that. But if so, that would constitute the entirety of the PR, which would at least allow for more time to assess the situation.
5. IF the version of Psych can’t be pegged, then I think we accept the breaking change and bump the semver appropriately.

Anyway, that’s my thoughts, FWIW :)

ethagnawl:
@jschank Which version of libyaml are you seeing this problem with?

If you don't happen to know, you should be able to use one of the following:
-  run this one-liner from your shell: `ruby -e 'require "psych"; v=Psych.libyaml_version; puts(v.join(", "))'`)
- `apt list --installed | grep libyaml`

jschank:
$ brew info libyaml reports version 0.2.5


one-line below seems to agree, returning 0, 2, 5


apt list won’t work - I am on a mac

ethagnawl:
Ah, okay. I thought you were on Ubuntu. For whatever it's worth, I'm not seeing the issue with libyaml 0.2.1 (Debian) or 0.2.2 (Ubuntu).

Anyways, I've taken a pass at deprecating the EOL'd Ruby versions that were causing the CI build to fail. Have a look at #823 and let me know if you have any feedback. Thanks for getting the ball rolling on this fix!

jschank:
Looks good to me.

Glad to have been helpful :)

ethagnawl:
Closed by #823.

pbek:
@ethagnawl, do you know when there will be a new release?

ethagnawl:
@pbek I just published [3.0.0](https://rubygems.org/gems/tmuxinator).

pbek:
Great, thank you. I will test it with the next release of https://github.com/pbek/QOwnNotes 😉

pbek:
I can confirm version 3.0.0 works again! Thank you very much!

PR #838, opened 2021-10-24
Simple maintenance improvements

a1346054:
Review comments are very welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,7 +8,7 @@ about: Create a bug report to help us improve tmuxinator
 A clear and concise description of what the bug is.
 
 **To Reproduce**
-Steps to reproduce the behavior. 
+Steps to reproduce the behavior.
 
 **Expected behavior**
 A clear and concise description of what you expected to happen.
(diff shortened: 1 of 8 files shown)


a1346054:
I can remove the alias-removing commit and create a new PR for it to discuss it there, are you ok with that?

The unnecessary shebang causes warnings in tools like lintian https://lintian.debian.org/tags/bash-completion-with-hashbang . A comment at the top that doesn't start with `#!` is perfectly fine to help guide text editors. I've also seen them have it as a comment at the very end of file (at least vim supports that). There are plenty of other alternatives with varying levels of support, such as having hints in .gitattributes, but ultimately it's up to the user to modify their own environment to understand what the file contains. The completion script is edited so very rarely that I don't see it as a big problem if the user has to type `:set ft=sh` or similar.

ethagnawl:
@a1346054 That all sounds good by me. Thanks again.

a1346054:
I removed the commit that removes the `mux` alias as requested, rebased on latest master to fix merge conflicts, and added a few fixes for typos.

ethagnawl:
Thanks, @a1346054! Would you mind adding an entry to the changelog for these changes? After that, I'll merge this PR.

akofink (approved):

Issue #841, opened 2021-12-13
gem env home changed by tmuxinator


hjw:
**Describe the bug**
Tmuxinator is changing my gem location.

If I start up a tmux session using tmux instead of tmuxinator I get:
```(ruby3.0.2)Portal/dhamma-forms-experiment(models✘) ↑ % gem env home
/Users/hjw/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0
```
This matches what I get from outside of a tmux session.

Within a tmux session started by tmuxinator my gem env home gets set to:
```
(ruby3.0.2)Portal/dhamma-forms-experiment(models✘) ↑ % gem env home
/usr/local/Cellar/tmuxinator/3.0.1/libexec
```

**To Reproduce**
from a plain old command line (outside of a tmuxinator session)
run `gem env home`
start a session using tmuxinator
run `gem env home`

**Expected behavior**
the path returned by `gem  env home`  within a tmuxinator started tmux session should match the path returned from outside of a tmuxinator started session.

**Environment**
tmuxinator 3.0.1
tmux 3.2a
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin18]
MacOS 10.14.6

stevehanson:
I hit this same issue after setting up a new machine last week and was able to get around it by uninstalling the Homebrew tmuxinator package and installing tmuxinator with Rubygems instead:

* Uninstall the Homebrew package: `brew remove tmuxinator`
* Exit out of tmux: `tmux kill-server`
* Start up tmux
* Install tmuxinator with Ruby: `gem install tmuxinator`

Everything now works for me. I'm not sure why the Homebrew package is not working correctly, but I wonder if the instructions should be removed from the readme or a warning added until this is fixed?

ethagnawl:
> I'm not sure why the Homebrew package is not working correctly, but I wonder if the instructions should be removed from the readme or a warning added until this is fixed?

Please feel free to submit a PR.

stevehanson:
Thanks, @ethagnawl. I opened #843.

I also wanted to provide some additional context:

* OS: macOS Monterey (12.0.1)
* Device: 2021 M1 Macbook Pro
* Architecture: I installed Homebrew in the normal way (non-Rosetta) and am not running my terminal with Rosetta
* Ruby: Ruby is installed with `rbenv`, not using system Ruby

Issue faced:

[...]

PR #843, opened 2021-12-17
Warn about Homebrew issues, recommend RubyGems

stevehanson:
Per https://github.com/tmuxinator/tmuxinator/issues/841, there might be issues when installing via HomeBrew. This PR adds a note warning that these issues exist and recommending that RubyGems be used instead of HomeBrew until the issues are resolved.


ethagnawl:
Thanks, @stevehanson!

Issue #846, opened 2022-01-13
docs(README.md):  the five standard layouts link is not working


halafi:
**Describe the bug**
see title

**To Reproduce**
click *the five standard layouts*

lyndsysimon:
This should be an easy fix. I'm looking for a good representation of the layouts now, and will submit a PR as soon as I find one that I like.

I'll update this comment as I find links that are "not quite right" for the docs but helpful if anyone else comes across this:

* https://dev.to/rajasegar/a-guide-to-create-tmux-like-custom-layouts-in-spacemacs-4j9p#layouts

lyndsysimon:
I looked up [the original link in the Wayback Machine](http://web.archive.org/web/20211001000000*/http://manpages.ubuntu.com/manpages/precise/en/man1/tmux.1.html), and it didn't include any graphical of what the layouts look like.

I'll try to find something better, but for now I'll just submit a PR that updates the link to an online version of the manpage to replace the one that rotted.

lyndsysimon:
PR submitted!

ethagnawl:
Closed via #854.

PR #847, opened 2022-01-27
Fix broken link in README (#846)

lyndsysimon:
I chose [this link](https://manpages.org/tmux/1) because it contains the same information as the previous page (as archived by the Wayback Machine from 2011 through 2021), in a clean and easy-to-read format.

If you'd like, I could also put together some screenshots of the various layouts and include it in a subsequent PR - but that seemed like overkill for this trivial change, and I wasn't sure you'd want to include the image files in this project's repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ windows:
 ```
 
 The layout setting gets handed down to tmux directly, so you can choose from
-one of [the five standard layouts](http://manpages.ubuntu.com/manpages/precise/en/man1/tmux.1.html#contenttoc6)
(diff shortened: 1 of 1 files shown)


ethagnawl:
Thanks for this, @lyndsysimon. I'm not familiar with manpages.org. Do you know if it's used by many other projects? Is it safe to assume that it's stable? On that note, I wouldn't be opposed to linking to the archived version of either page.

ethagnawl:
Closed by #854.

PR #859, opened 2022-06-02
allow tmux 3.3

mstruebing:
I'm not sure if this is everything which is needed. If not, please let me know what else I need to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rvm:
   - "2.7.3"
   - "3.0.1"
 env:
+  - TMUX_VERSION=3.3
   - TMUX_VERSION=3.2a
   - TMUX_VERSION=3.2
   - TMUX_VERSION=3.1c
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 ### Misc
+- add tmux 3.3 to (currently defunct) Travis test matrix; add 3.3 to supported tmux versions list
 - use stable tmux links in README.md
 
 ## 3.0.2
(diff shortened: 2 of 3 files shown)


ethagnawl:
Thanks, @mstruebing. I've updated the changelog and verified that the test suite passes using tmux 3.3. I'll cut a release shortly.

Frederick888:
It seems that it has to be `3.3` (without quotes)? I'm using ruby 3.0.4p208 (2022-04-12 revision 3fa771dded).

mstruebing:
Hey @Frederick888 - why do you think that?

Frederick888:
@mstruebing 

```
$ tmux -V
tmux 3.3
$ tmuxinator start default
    WARNING: You are running tmuxinator with an unsupported version of tmux.
    Please consider using a supported version:
    (3.3, 3.2a, 3.2, 3.1c, 3.1b, 3.1a, 3.1, 3.0a, 3.0, 2.9a, 2.9, 2.8, 2.7, 2.6, 2.5, 2.4, 2.3, 2.2, 2.1, 2.0, 1.9, 1.8, 1.7, 1.6, 1.5)

Press ENTER to continue.
```

It stopped showing this warning after I removed the quotes.

mstruebing:
I'm not a ruby person but I would be very surprised if everything is ok with quotes but that one item isn't.

Frederick888:
@mstruebing 
https://github.com/tmuxinator/tmuxinator/blob/5f42cc9b44587ccdf0e27a45095bbb475d815e1c/lib/tmuxinator/tmux_version.rb#L36-L38
https://github.com/tmuxinator/tmuxinator/blob/5f42cc9b44587ccdf0e27a45095bbb475d815e1c/lib/tmuxinator/config.rb#L63-L73

```
irb(main):003:0> [3.3].include? "3.3".to_s[/\d+(?:\.\d+)?/, 0].to_f
=> true
irb(main):004:0> ["3.3"].include? "3.3".to_s[/\d+(?:\.\d+)?/, 0].to_f
=> false
```

Issue #861, opened 2022-06-02
Does not work with tmux 3.3


CaledoniaProject:
**Describe the bug**
It no longer runs

```
%> tmux -V
tmux 3.3

%> tmuxinator start xxx
    WARNING: You are running tmuxinator with an unsupported version of tmux.
    Please consider using a supported version:
    (3.3, 3.2a, 3.2, 3.1c, 3.1b, 3.1a, 3.1, 3.0a, 3.0, 2.9a, 2.9, 2.8, 2.7, 2.6, 2.5, 2.4, 2.3, 2.2, 2.1, 2.0, 1.9, 1.8, 1.7, 1.6, 1.5)

Press ENTER to continue.
server exited unexpectedly
no server running on /private/tmp/tmux-501/default
no server running on /private/tmp/tmux-501/default
no server running on /private/tmp/tmux-501/default
no sessions
```

**Environment**
 - tmuxinator Version: muxinator 3.0.3
- tmux Version: tmux 3.3
 - Ruby Version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin21]
 - OS: Mac 12.4

tnzk:
As another data point: I've got the same warning but it works fine.

- tmuxinator Version: muxinator 3.0.2
- tmux Version: tmux 3.3
- Ruby Version: ruby 2.6.8p205 (2021-07-07 revision 67951) [universal.arm64e-darwin21]
- OS: Mac 12.4

azhargiri:
![image](https://user-images.githubusercontent.com/904684/171808395-6d60eaf8-7964-45c4-a923-75c23ddbb637.png)

same here

tmux 3.3
tmuxinator 3.0.3
ruby 3.1.2
OS: Archlinux

PR #862, opened 2022-06-03
fix tmux 3.3 version

azhargiri:
Fix issue #861 

Run tmuxinator on tmux 3.3 show warning that version is unsupported. This commit fix tmux version declared inside tmux_version to not use string.


ethagnawl:
Thanks, @azhargiri! I'd actually fixed this separately before I saw your PR. A new release should be available momentarily.

azhargiri:
Thanks @ethagnawl 

happy to hear that

PR #872, opened 2022-12-15
Check for github action updates using dependabot

jlosito:
This will use dependabot to check for any updates to the github action packages which this project uses and submit pull requests with version bumps in order to keep packages up-to-date on a monthly basis.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


akofink (approved):
This would be an excellent improvement IMO 🥇

akofink:
This doesn't feel like it needs a CHANGELOG entry. Any objections @tmuxinator/tmuxinator?

PR #880, opened 2023-10-14
Remove `mux` alias in Bash completion file

### LINKED ISSUE #868: Use of `alias` in the completion file
**Describe the bug**

I noticed that in `tmuxinator.bash`, there is a line:

```sh
alias mux="tmuxinator"
```

and also in zsh:

```sh
abbr --add mux "tmuxinator"
```

**Expected behavior**
I expect no completion file to automatically create any aliases - It's convention not to, and is kind of sneaky, because it's not really expected. It is also inconstant because there is no such alias in the Fish completions file. Please remove this? It is a breaking change, but of the people that actually use this alias can add it to their own bashrc or zshrc

**Environment**
- Any Bash environment

hyperupcall:
It should be removed because:

- Aliases in completion files are unconventional
- This will conflict with user-defined aliases in unexpected and hard-to-debug ways
- It is not consistent with completions for the other shells that we support (zsh)
- Is confusing

Closes #868

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 ### Misc
 - Fix typos and remove extra whitespace
+- Remove `mux` alias from Fish and Bash completions
 
 ## 3.1.0
 - add a FAQ entry about how long commands may be lost/corrupted by TTY typeahead
--- a/completion/tmuxinator.bash
+++ b/completion/tmuxinator.bash
@@ -19,5 +19,4 @@ _tmuxinator() {
     fi
 }
 
-complete -F _tmuxinator tmuxinator mux
-alias mux="tmuxinator"
+complete -F _tmuxinator tmuxinator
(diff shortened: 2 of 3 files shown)


akofink (approved):
Makes sense to me and matches the other shell completion files, as you point out. I personally do have this alias explicitly defined in my `.zshrc`.

akofink:
Would you mind adding an entry to the `CHANGELOG.md`?

We could also probably remove this from `tmuxinator.fish`.

hyperupcall:
@akofink Good idea, I've added an entry to the changelog and removed it from Fish as well

akofink (approved):

akofink:
Any other feedback from @tmuxinator/tmuxinator on this one?
The change looks ready to merge to me, but I can't say I'm an expert on these shell completions.

akofink:
Hey there, apologies for not merging this earlier!

Would you mind addressing the conflicts in the CHANGELOG.md? Then we should be good to merge. 🙏

hyperupcall:
@akofink No worries, addressed the merge conflicts

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Misc
 - Fix typos and remove extra whitespace
 - Set pane title if provided in config file
+- Remove `mux` alias from Fish and Bash completions
```
akofink, line 8: This change should be under the `## Unreleased` header, above 3.1.2.

akofink:
This was merged as https://github.com/tmuxinator/tmuxinator/pull/906

Thanks again, @hyperupcall!

PR #882, opened 2023-12-31
Change thor version in gemspec >=1.2.1

Dstracke:
Change gemspecs to repair this bug:
/usr/lib/ruby/3.0.0/rubygems/specification.rb:1453:in `rescue in block in activate_dependencies': Could not find 'thor' (~> 1.2.1)


ethagnawl:
Thanks for this, @Dstracke.

I think we should prefer a "pessimistic" constraint which should (in an ideal world) prevent future compatibility issues with thor. 

The version shared by @lervag in #883 should work. Do you want to do the honors? If not, I can update this PR or open a new one.

Dstracke:
I can absolutely understand the decision.
It is perfectly fine for me if the PR is updated or a new one is created.

lervag:
Will you update your PR, @Dstracke?

PR #788, opened 2020-09-27
Fixes installation of a completion file for fish

satoryu:
The command in README in order to install a completion file for fish doesn't work. 
It looks need `-O` option as well as other examples.

PR #888, opened 2024-01-24
docs: fish completions file not downloaded to correct directory

geoffreyvanwyk:
The command given in the README.md downloads the fish file to the current working directory instead of to the provided directory, because the `-O` option is left out. Additionally, the `-O` option's value must a file, not a directory.

Having the `$` sign at the start of the given command is inconvenient, because, after copying and pasting, one has to firstly delete it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Note: ZSH's completion files can be put in other locations in your `$fpath`. Ple
 
 ### fish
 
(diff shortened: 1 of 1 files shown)


akofink (approved):
Makes sense to me 👍

akofink:
This is essentially a duplicate of https://github.com/tmuxinator/tmuxinator/pull/788

akofink:
This change was merged as https://github.com/tmuxinator/tmuxinator/pull/788 Thanks again!

PR #891, opened 2024-02-02
Set pane title if provided in config file

augustelalande:
Pane titles are a commonly desired feature in tmux sessions (e.g. https://stackoverflow.com/questions/47800955/how-to-set-pane-titles-with-tmuxinator), but not explicitly supported by tmuxinator (although a pane can be named).

To pass a title to tmux it is generally required to use a somewhat hacky method, and run the command:
`printf '\033]2;%s\033\\' '<title>'` (`tmux select-pane -T <title>` is also supported).

Since tmuxinator already provides the ability to define a pane title, I thought I would take it a step further and pass that title to tmux. As it stands right now, tmux would still need to be configured with the appropriate settings to display the title, e.g.:
```
set -g pane-border-format "#{pane_index} #{pane_title}"
set -g pane-border-status bottom
```
but at least the title will be passed directly from the tmuxinator config, to tmux. As a follow up to this PR, we could make the pane title configuration part of the tmuxinator config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## 3.1.0
 - add a FAQ entry about how long commands may be lost/corrupted by TTY typeahead
 - increment thor minor version in tmuxinator.gemspec
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.2.0
+- Set pane title if provided in config file
+
 ## 3.1.0
 - add a FAQ entry about how long commands may be lost/corrupted by TTY typeahead
 - increment thor minor version in tmuxinator.gemspec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 3.2.0
+## Unreleased
 - Set pane title if provided in config file
 
 ## 3.1.0
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ for a workaround.
 
 ### Pane titles
 
-It is also possible to give a title to panes.
+It is also possible (starting with tmux v2.6) to give a title to panes.
 
 ```yaml
 windows:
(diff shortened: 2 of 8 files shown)


ethagnawl:
This is great, @augustelalande!

A few initial bits of feedback:
- It'd be great to get some test coverage around these changes
- Should template.rb detect incompatible versions of tmux and print a warning?
- This will need to be documented in the README and it might also be worth adding to the default template
- Please update the CHANGELOG :smile_cat: 

Do you have any thoughts, @akofink?

akofink (commented):
The changes look good to me, and it seems to work as expected. Very cool improvement! 💯 
I agree with all of @ethagnawl's points as well. It looks like only tmux 2.6 and above will support this, [from here](https://raw.githubusercontent.com/tmux/tmux/master/CHANGES):

```
CHANGES FROM 2.5 TO 2.6, 05 October 2017

* Add select-pane -T to set pane title.
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 3.2.0
```
akofink, line 1: I believe this should be `## Unreleased` for now.

https://github.com/tmuxinator/tmuxinator/blob/master/CONTRIBUTING.md?plain=1#L30

augustelalande: thanks fixed

augustelalande:
Thanks for the feedback guys. I think I've implemented all the changes, with two caveats
- I check tmux version but don't print a warning. I'm not sure it should print a warning since right now setting the title won't do anything unless explicitly configured in tmux settings as stated above, so it would be weird to print a warning for a feature which is "disabled" by default.
- For the same reason I didn't add it to the default template since it needs additional configuration to work.

My proposal would be once this gets merged, to add configuration options to the tmuxinator config, to
- enable pane titles
- configure pane title position
- configure pane title format

akofink (approved):
Looks ready to merge to me. Nice work! The tests look great. 🙌

ethagnawl:
My one remaining bit of feedback is that the README should specify the minimum version of tmux required to make use of this functionality. 

Otherwise, LGTM. Thanks again, @augustelalande!

augustelalande:
> My one remaining bit of feedback is that the README should specify the minimum version of tmux required to make use of this functionality.
> 
> Otherwise, LGTM. Thanks again, @augustelalande!
Done

akofink:
Thanks for this awesome improvement @augustelalande! 🦖

Issue #884, opened 2024-01-08
Call for Maintainers


ethagnawl:
I have been the primary maintainer of this library for the last few years but I no longer have the bandwidth to do so.

I'm planning to cut one final release (3.1.0) and then turn maintenance back over to @tmuxinator/tmuxinator. It'd also be great to onboard new folks who'd be interested in contributing. I'm not going to disappear entirely and abandon the project but I can't promise that I'll be triaging bug reports, cutting releases, etc. with any sort of regularity going forward.

dysnomian:
I've been looking for an OSS project to get involved with and I love tmuxinator! I'll email you.

akofink:
I'd be happy to get involved. I've been using tmuxinator for about 12 years, and I still use it every day. I'd like to see the project continue to improve incrementally and maintain support for newer releases of Ruby and tmux.

ethagnawl:
@geoffreyvanwyk Yes. A number of wonderful folks have gotten involved since this issue was initially opened and have taken the reins. If you'd like to get involved, please connect with @akofink.

PR #892, opened 2024-02-08
Add configuration options for pane titles

augustelalande:
Hi guys, as a follow up to my previous pull request (#891), I want to add configuration options to enable pane titles in tmuxinator without having to mess with tmux.conf. Before I go further with the development I wanted to get your feedback on what I have so far, and whether it fits with the tmuxinator mission. I have added 3 configuration options:

1. `enable_pane_titles`  # Enables pane titles by setting a default configuration for `pane-border-status` and `pane-border-format`
2. `pane_title_position`  # Overwrites the default `pane-border-status` with the users choice
3. `pane_title_format`  # Overwrites the default `pane-border-format` with the users choice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 ### Fixes
 - Shell escape pane titles to fix multi word and special character titles
+- Add configuration options for pane titles
 
 ## 3.1.2
 ### Fixes
--- a/README.md
+++ b/README.md
@@ -151,10 +151,11 @@ root: ~/
 # Enables the display of pane titles. For example "editor" below. Defaults to false.
 # enable_pane_titles: true
 
-# Configures pane title position. Either, bottom or top. Defaults to top.
+# Configures pane title position. Can be: bottom, top, or off. Defaults to top.
 # pane_title_position: bottom
 
 # Configures pane title format. Defaults to "#{pane_index}: #{pane_title}".
+# Please see the tmux manpage for details, on valid formats.
 # pane_title_format: " [ #T ] "
 
 windows:
--- a/README.md
+++ b/README.md
@@ -148,6 +148,15 @@ root: ~/
 # Controls whether the tmux session should be attached to automatically. Defaults to true.
 # attach: false
 
+# Enables the display of pane titles. For example "editor" below. Defaults to false.
+# enable_pane_titles: true
+
+# Configures pane title position. Either, bottom or top. Defaults to top.
+# pane_title_position: bottom
+
+# Configures pane title format. Defaults to "#{pane_index}: #{pane_title}".
+# pane_title_format: " [ #T ] "
+
 windows:
   - editor:
       layout: main-vertical
@@ -155,7 +164,8 @@ windows:
       # 'before' represents legacy functionality and will be deprecated in a future release, in favour of 'after'
       # synchronize: after
       panes:
-        - vim
+        - editor:
+          - vim
         - guard
   - server: bundle exec rails s
   - logs: tail -f log/development.log
@@ -222,6 +232,8 @@ for a workaround.
 It is also possible (starting with tmux v2.6) to give a title to panes.
 
 ```yaml
+enable_pane_titles: true
+
 windows:
   - editor:
       layout: main-vertical
@@ -229,13 +241,6 @@ windows:
         - editor:
           - vim
         - guard
-```
-
-**Note:** For the titles to display you will need to modify your .tmux.conf with the following entries.
-
-```
-set -g pane-border-format "#{pane_index} #{pane_title}"
-set -g pane-border-status bottom
 ```
 
 ## Interpreter Managers & Environment Variables
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ root: ~/
 # Enables the display of pane titles. For example "editor" below. Defaults to false.
 # enable_pane_titles: true
 
-# Configures pane title position. Can be: bottom, top, or off. Defaults to top.
+# Configures pane title position. Can be: bottom, top, or "off". Note: "off" must be provided in quotes to avoid being interpreted as a boolean false. Defaults to top.
 # pane_title_position: bottom
 
 # Configures pane title format. Defaults to "#{pane_index}: #{pane_title}".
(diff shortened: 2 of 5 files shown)


akofink:
Hey @augustelalande - thanks for another useful enhancement! Side note: curious if you saw https://github.com/tmuxinator/tmuxinator/issues/884 😉 

I think this change makes sense for someone who wants pane titles in only some of their projects but doesn't want to set it in their global tmux config. This seems quite useful to me, since only some of my projects have pane titles configured after your most recent PR.

I've only recently started contributing to tmux, but this idea of configuring tmux per project seems to already exist, like with the [`tmux_options` project config](https://github.com/tmuxinator/tmuxinator/blob/ba62c5d4ecfa2b7b4521bbf58b08a5b5a0a419ab/README.md?plain=1#L137), so it seems to fit the spirit of tmuxinator.

We'd want to document this somewhere in the readme, maybe under the new pane titles section in addition to the project config in the readme. We could also update the [sample project config](https://github.com/tmuxinator/tmuxinator/blob/ba62c5d4ecfa2b7b4521bbf58b08a5b5a0a419ab/spec/fixtures/sample.yml).

Any thoughts from others in @tmuxinator/tmuxinator?

augustelalande:
Hey @akofink appreciate it. I did see that post but I don't think I have much time to consistently contribute at the moment, I'm very much swamped at work. On the same note, I want to share my appreciation for you and all the OSS community, I'll keep doing my small part, and contribute where I can.

My only concern regarding the tmux.conf is that in that case the configuration is still "handled" by tmux itself, whereas here tmuxinator would be taking over some of that responsibility. It's not necessarily a big deal, but just thought I would mention it.

I will add the documentation and test cases once we've all agreed this should proceed.

Has @ethagnawl officially handed the torch to you?

akofink:
I thought it'd be a great idea to help more with OSS that I use like this project because I am also busy/motivated at work and often tempted to work after hours. This helps me put that energy to good use without deteriorating work-life balance. It sounds a bit counter intuitive now that I've typed it out 😆 

> Has @ethagnawl officially handed the torch to you?

I don't really see it that way. I'm hopeful we can continue to work together as a group of maintainers. I don't always have time to stay on top of it all the time. I'm hopeful we can gain other interested folks who are also willing to participate. If you ever change your mind, we'll still be here :)

lib/tmuxinator/project.rb
```diff
@@ -381,5 +381,43 @@ def parsed_parameters(parameters)
...
+
+    def pane_title_position?
+      yaml["pane_title_position"]
+    end
+
+    def tmux_set_pane_title_format
+      if pane_title_format?
+        "#{tmux} set pane-border-format \"#{yaml['pane_title_format']}\""
+      else
+        "#{tmux} set pane-border-format " + '"#{pane_index}: #{pane_title}"'
```
akofink, line 405: I haven't tested this, but [code climate is complaining about the interpolation within single quotes](https://codeclimate.com/github/tmuxinator/tmuxinator/pull/892), which makes sense!

augustelalande: Ya that can't be helped since tmux use the same text formating as ruby. Unless you know a way?

ethagnawl: You could try "interpolable" string literals: https://docs.ruby-lang.org/en/master/syntax/literals_rdoc.html#label-25+and+-25Q-3A+Interpolable+String+Literals

augustelalande: hmm, not sure what you mean. I attempted it in my latest commit, but it's still rejected. Maybe I did it wrong though.

augustelalande: ok figured something out

augustelalande:
Should be ready for proper review now

augustelalande:
@akofink Is anything blocking this?

README.md
```diff
@@ -148,14 +148,25 @@ root: ~/
 # Controls whether the tmux session should be attached to automatically. Defaults to true.
 # attach: false
 
+# Enables the display of pane titles. For example "editor" below. Defaults to false.
+# enable_pane_titles: true
+
+# Configures pane title position. Can be: bottom, top, or off. Defaults to top.
+# pane_title_position: bottom
```
akofink, line 155: I tried setting this to `pane_title_position: off` while keeping `enable_pane_titles: true`, and I still see pane titles. I wonder if you could try it and confirm the behavior?

akofink (commented):
I tested it out, and it works mostly as expected - just one edge case that I have a question about inline. Other than that, I think it's ready for merge. 💯

README.md
augustelalande: Oh ya, it's because `off` is actually a yaml keyword in [yaml 1.1](https://yaml.org/type/bool.html) which is interpreted as a boolean false. If you want to pass the off string, you have to explicity quote it `"off"`.

I'm not sure how we should address this here. We could just say we don't officially support off, since it's not very useful, and will lead to confusion.

augustelalande: Or we could re-interpret a boolean false as an "off" string.

akofink: Makes sense! I think it's probably fine as is - just a small yaml gotcha.

Another case I thought of is if the user has pane titles enabled in their `.tmux.conf`, but `enable_pane_titles: false` in the tmuxinator project yaml, the pane titles still show in that session as it's currently written. That said, since the default is `false`, it wouldn't be a great user experience for pane titles to be turned off if you're expecting them on as defined in your `.tmux.conf`.

augustelalande: Ya I think it's fine for .tmux.conf to take precedence.

augustelalande: I will add a comment about the `off` thing just to try and reduce confusion.

akofink (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## Unreleased
-### Fixes
 - Shell escape pane titles to fix multi word and special character titles
+- Add configuration options for pane titles
```
akofink, line 3: Would you mind rebasing with `master` to fix up the changelog?

augustelalande: Done

akofink:
Thanks again, @augustelalande! 🥇

PR #896, opened 2024-02-14
feat: Add Ruby 3.2 to the test matrix

akofink:
This PR is mostly to demonstrate the issue solved by https://github.com/tmuxinator/tmuxinator/pull/889

#### For further discussion

It's sort of unclear which versions of Ruby we want to test against at the moment. Based on the [ruby maintenance schedule](https://www.ruby-lang.org/en/downloads/branches/), Ruby < 3.0 was end of life as of 2023-03-31.

I'm open to the idea of cleaning up the test matrix and removing Ruby 2 support for tmuxinator.

Ruby 3.0 is also going end of life quite soon, 2024-03-31. How long should we continue testing with it after EOL?

Should we also maybe test against Ruby 3.1?

We're also testing against literally all versions of tmux since 1.5.
Would it make sense to at least collapse `3.1[,a,b,c]` to just `3.1c`?

Happy to report an issue for this discussion about which versions we want to support and test against.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -21,6 +21,7 @@ engines:
     enabled: false
   rubocop:
     enabled: true
+    channel: rubocop-0-61-1
 ratings:
   paths:
   - Gemfile.lock
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,7 @@
+# Run this command to always ignore formatting commits in `git blame`
+# git config -g blame.ignoreRevsFile .git-blame-ignore-revs
+
+# chore: Fix rubocop failures
+# Author: Andrew Kofink <ajkofink@gmail.com>
+# Date:   Sat Mar 30 12:13:30 2024 -0400
+1aee1bc3615e431093e6301020572b16b89dfbae
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,6 +11,8 @@ jobs:
       fail-fast: false
       matrix:
         ruby_version:
+          - '3.2.3'
+          - '3.1.4'
           - '3.0.1'
           - '2.7.3'
           - '2.6.7'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,20 +1,13 @@
 AllCops:
-  Include:
-  - "**/*.podspec"
-  - "**/*.jbuilder"
-  - "**/*.opal"
-  - "**/Vagrantfile"
-  - "**/Berksfile"
-  - "**/Cheffile"
-  - "**/Vagabondfile"
   Exclude:
   - "vendor/**/*"
   - "db/schema.rb"
   DisplayCopNames: true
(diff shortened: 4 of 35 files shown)


augustelalande:
I think dropping <3.0 makes sense, and is in line with the policy stated in the README
> tmuxinator aims to be compatible with the currently maintained versions of Ruby

By the same logic we could drop 3.0 as of the end of the month, but assuming it's not blocking anything else, you could keep it an extra year until April 2025.

I think 3.1 should be added to keep in line with the policy and possibly also 3.3?

augustelalande:
Is there any reason you want to reduce the number of tmux versions? Is backwards compatibility blocking anything?

akofink:
The introduction of Ruby > 3.0 causes rubocop to fail. [I've attempted upgrading](https://github.com/tmuxinator/tmuxinator/compare/master...akofink:tmuxinator:update-rubocop); it's quite a big task. I think we have to get to [v0.61.0](https://github.com/rubocop/rubocop/releases/tag/v0.61.0) to get [this fix](https://github.com/rubocop/rubocop/commit/a639a175ab80a53864498a59d04dfcc1d3249d9c).

> Is there any reason you want to reduce the number of tmux versions?

We're currently testing against 3.1, 3.1a, 3.1b, and 3.1c; tmux appears to follow [semver](https://semver.org) conventions, where there are expected to be no breaking changes between 3.1 and 3.1c. I don't really feel that strongly about reducing the tmux versions list if there's a good reason to test against each version like this.

augustelalande:
Nice. Are you running into any blockers when upgrading rubocop?

augustelalande:
> We're currently testing against 3.1, 3.1a, 3.1b, and 3.1c; tmux appears to follow [semver](https://semver.org/) conventions, where there are expected to be no breaking changes between 3.1 and 3.1c. I don't really feel that strongly about reducing the tmux versions list if there's a good reason to test against each version like this.
Like I said, I don't really see a point in dropping any of the versions even if they don't do anything. At worst it's just a bit of wasted compute on the microsoft servers.

akofink:
Rubocop adds cops as it upgrades, and we have some failing cops. [One way to upgrade is to rely on `.rubocop_todo.yml`](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration) which just excludes existing occurrences of failures, which we can work on over time to correct.

It also seems like upgrades between major versions of rubocop (i.e. 0.x -> 1.x -> 2.x) have more than just new cops. Many cops are renamed, split, or merged into other cops. I'll open a PR at some point when I get it to a good state, but no promises on timeline. 😄

augustelalande:
Ok. Let me know if you need help on anything.

akofink:
I've rebased this PR on top of https://github.com/tmuxinator/tmuxinator/pull/905 to demonstrate rubocop works with all existing supported versions of ruby in addition to 3.1 and 3.2.

akofink:
Closing this in favor of https://github.com/tmuxinator/tmuxinator/pull/907

PR #898, opened 2024-02-15
Update tmux_version.rb

robplan:
3.4 should be treated as a float because it doesn't have a letter in the version number. 
Making it a string makes `supported?` return false instead of true for tmux v 3.4
image (8)


akofink (approved):
Thank you! Apologies for this 🙈

PR #902, opened 2024-02-17
fix: Shellescape pane titles

### LINKED ISSUE #900: Pane Titles Bugs (Parsing + Not Setting Multiple)
**Parsing errors occur in providing special characters as well as setting multiple panes**
The recent feature to automatically set the pane title is extremely welcome, but currently has two notable bugs.

The first is when special characters, such as `(` or `)` are set in the pane title:
image

You will receive an error saying:
``` bash
sh: 37: Syntax error: "(" unexpected
```

The second notable error is that multiple pane titles are not parsed correctly.
Config:
image
Session:
image
Error output:
image

It is somewhat interesting to note that the output is always less than the full list, suggesting that each pane title was attempted to be set but failed except the first.

`select-pane` to set the title manually still works (although the error output is still seen from the tmuxinator parsing).
Example config file:
image
Example of created session using `select-pane`
image

**To Reproduce**
Create a config YAML file with a title using `(` or multiple panes.

**Environment**
 - tmuxinator Version: tmuxinator 3.1.1
- tmux Version: tmux 3.3a
 - Ruby Version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
 - OS: image

**Additional information**
Although primary testing was in Kali/Debian, it was noted that Ubuntu 20.04 had identical errors. 
OS details:
image

**Fixes**
I have not delved very deep into the Tmuxinator codebase, but my best assumption for the multiple pane issue is that the target defined below is not set correctly.
image

For the parsing issue, I have a suspicion that simply wrapped the target in single quotes would work to solve this problem.

akofink:
#### Problem
Pane titles are currently not shell escaped, so multi word titles or titles containing shell characters break, i.e.

```
$ mux pane-names
usage: select-pane [-DdeLlMmRUZ] [-T title] [-t target-pane]
sh: 36: Syntax error: "(" unexpected
```

Session and window names are already escaped in this way.

```
name: pane-names (a test! | let's find out)
root: ~/dev/sandbox

windows:
  - editor (and others):
      layout: tiled
      panes:
        - vim and things:
          - vim
        - mostly just docker (maybe compose):
          - docker ps
```

Fixes #900

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## Unreleased
+### Fixes
+- Shell escape pane titles to fix multi word and special character titles
 
 ## 3.1.2
 ### Fixes
--- a/lib/tmuxinator/pane.rb
+++ b/lib/tmuxinator/pane.rb
@@ -7,7 +7,7 @@ module Tmuxinator
       @index = index
       @project = project
       @tab = tab
-      @title = title
+      @title = title.to_s.shellescape unless title.nil?
     end
 
     def tmux_window_and_pane_target
--- a/spec/lib/tmuxinator/pane_spec.rb
+++ b/spec/lib/tmuxinator/pane_spec.rb
@@ -10,7 +10,7 @@ describe Tmuxinator::Pane do
   let(:project) { double }
   let(:window) { double }
   let(:commands) { ["vim", "bash"] }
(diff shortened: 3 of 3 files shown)


augustelalande:
Is anything holding this up?

PR #905, opened 2024-03-30
chore: Update rubocop to 0.61 for Ruby > 3.0 support

akofink:
## Problem

See rubocop error (not just style failures) on this action run that adds Ruby 3.2 to the test matrix: https://github.com/tmuxinator/tmuxinator/actions/runs/8307640599

We have to upgrade rubocop to [v0.61.0](https://github.com/rubocop/rubocop/releases/tag/v0.61.0) to get [this fix](https://github.com/rubocop/rubocop/commit/a639a175ab80a53864498a59d04dfcc1d3249d9c) related to yaml parsing.

## Solution

Update rubocop in the gemspec, and correct changes to cops between 0.48.1 and 0.61.1.

Add a `.git-blame-ignore-revs` file so our blames don't include this massive rubocop auto-fix.

## To test

Test against existing supported versions (also done in our CI actions):

```sh
$ rbenv shell 3.0.1
$ bundle exec rake rubocop
```

Test against Ruby > 3.0. On master, this will break:

```sh
$ rbenv shell 3.1.4
$ bundle exec rake rubocop
```

### DIFF
--- a/spec/lib/tmuxinator/project_spec.rb
+++ b/spec/lib/tmuxinator/project_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "spec_helper"
 
 describe Tmuxinator::Project do
@@ -589,7 +591,7 @@ describe Tmuxinator::Project do
   end
 
   describe "::load" do
-    let(:path) { File.expand_path("../../../fixtures/sample.yml", __FILE__) }
+    let(:path) { File.expand_path("../../fixtures/sample.yml", __dir__) }
     let(:options) { {} }
 
     it "should raise if the project file doesn't parse" do
@@ -692,7 +694,7 @@ describe Tmuxinator::Project do
 
       it "configures a default format" do
         expect(project.tmux_set_pane_title_format).to eq(
-          'tmux set pane-border-format "#{pane_index}: #{pane_title}"'
+          "tmux set pane-border-format \"\#{pane_index}: \#{pane_title}\""
         )
       end
     end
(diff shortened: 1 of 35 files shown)


akofink:
@augustelalande this should be good to go now. You can also see it works with Ruby 3.1 and 3.2 on  https://github.com/tmuxinator/tmuxinator/pull/896. Any thoughts/suggestions? 🙏

spec/lib/tmuxinator/project_spec.rb
```diff
@@ -691,9 +693,11 @@
       before { project.yaml["pane_title_format"] = nil }
 
       it "configures a default format" do
+        # rubocop:disable Lint/InterpolationCheck
         expect(project.tmux_set_pane_title_format).to eq(
           'tmux set pane-border-format "#{pane_index}: #{pane_title}"'
```
augustelalande, line 698: I think you can appease rubocop by replacing the line with
```suggestion
          "tmux set pane-border-format \"\#{pane_index}: \#{pane_title}\""
```
kinda ugly, but also makes it clear the non-interpolation is intentional.

augustelalande (approved):
Nice job on the clean up

akofink:
I merged this with a merge commit so the .git-blame-ignore-revs would work properly (i.e. the commit SHA wouldn't change), and it seems to have worked. See this line changed with the massive rubocop fix in the blame still points to the prior change on that line from 9 years ago: https://github.com/tmuxinator/tmuxinator/blame/master/lib/tmuxinator/cli.rb#L198

🎉

PR #908, opened 2024-03-31
feat: Add support for tmuxinator append

akofink:
## Metadata

Original PR from @flovilmart: https://github.com/tmuxinator/tmuxinator/pull/652

## Problem

Closes: https://github.com/tmuxinator/tmuxinator/issues/265

It often makes sense to want to add a tmuxinator project to an existing tmux session vs. creating a new session.

## Solution

Add a new option to append a project's windows to the current session instead of creating a new session.

```
tmuxinator start [project] --append
```

## Testing

I've covered the code additions with additional unit tests where the behavior changes when being called with `--append`.

### Relevant Scenarios

Using `--append` outside a tmux session results in an error.

```sh
$ (cd ~/dev/repos/tmuxinator; bundle exec tmuxinator start --append 908)
Cannot append to a session that does not exist
```

Test that `--append` captures the correct next window index by running it from sessions of various numbers of existing windows:

```sh
# From within tmux session "908"
$ (cd ~/dev/repos/tmuxinator; bundle exec tmuxinator start --append 908)

# Also try it across session names, like from a session named "test"
$ (cd ~/dev/repos/tmuxinator; bundle exec tmuxinator start --append 908)
```

The first example results in a tmux session with windows like this, where the "908" project has windows named `window1`, `window2`, `window3`. Note the window indexes as well:

Screenshot 2025-03-16 at 12 35 14

The second example shows appended windows of different names than the existing current session windows:

Screenshot 2025-03-16 at 12 55 28

You can append as many times as you'd like until you hit limits with the number of tmux windows per session.

## Original Description / Commit Messages

Adds support for tmuxinator append

Adds ability to specify default loading behavior (start or append)

- Use start when not possible to append

nits

nits and improvements

refactor Cli::bootstrap (#1)

Adds reference to the new command

use say

cli: Adds tests for bootstraping with option set to append

adds tests for project window index

refactor to use start --append instead of append. improve config loading

reduce complexity of the start method

reduce complexity in cli.rb

void(commit)

extract get params

Update lib/tmuxinator/config.rb

Update lib/tmuxinator/project.rb

Update lib/tmuxinator/cli.rb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## Unreleased
+### Features
+- Add support for tmuxinator start --append
 
 ## 3.3.3
 ### Features
--- a/README.md
+++ b/README.md
@@ -398,6 +398,11 @@ Shows tmuxinator's version.
 tmuxinator version
 ```
 
+Append a project's windows to the current session (instead of creating a new session)
+```
+tmuxinator start [project] --append
+```
+
 ## Project Configuration Location
 
 Using environment variables, it's possible to define which directory
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -1,17 +1,19 @@
 #!<%= ENV["SHELL"] || "/bin/bash" %>
 
-# Clear rbenv variables before starting tmux
-unset RBENV_VERSION
-unset RBENV_DIR
+<%- if !append? -%>
+  # Clear rbenv variables before starting tmux
+  unset RBENV_VERSION
+  unset RBENV_DIR
 
-<%= tmux %> start-server;
+  <%= tmux %> start-server;
+<%- end -%>
 
 cd <%= root || "." %>
 
 # Run on_project_start command.
 <%= hook_on_project_start %>
 
-<%- if !tmux_has_session? name -%>
+<%- if append? || !tmux_has_session?(name) -%>
 
   # Run pre command.
   <%= pre %>
@@ -98,7 +100,7 @@ cd <%= root || "." %>
   <%= hook_on_project_restart %>
 <%- end -%>
 
-<%- if attach? -%>
+<%- if attach? && !append? -%>
   if [ -z "$TMUX" ]; then
     <%= tmux %> -u attach-session -t <%= name %>
   else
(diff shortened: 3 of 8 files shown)


lib/tmuxinator/config.rb
```diff
@@ -63,6 +67,12 @@ def default
         "#{directory}/default.yml"
       end
 
+      def options
+        YAML.safe_load(File.read(options_file), aliases: true) || {}
```
akofink, line 71: This needed updating for https://github.com/ruby/psych/issues/533

martin-braun:
Does this run `pre_window` on those windows that are appended? If so, is there maybe a way to tell it not to run `pre_window` when appending?

hschne:
@akofink I Usually have a bunch of existing sessions that I want to add some windows to, and this would make my life so much easier. Any idea when this will be merged/released? :slightly_smiling_face: 

I'd be more than happy to lend a hand to move this forward, just need to know what needs doing :)

akofink:
Hi @hschne, thanks for your interest in helping get this across the line!

I've updated the branch, resolving conflicts and cleaning up some of the extra functionality that wasn't directly related to this new CLI option. If you could give it a test, that would be great - to test locally, just run it via `bundle exec`:

```sh
cd ~/dev/repos/tmuxinator

# Ensure you're on some supported version of ruby, i.e.
# rbenv shell 3.3.7

bundle install
bundle exec tmuxinator debug --append 908
bundle exec tmuxinator start --append 908
```

The debug output should be the exact script the `start` command will run in the context of the current shell, in case you see weird behavior.

akofink:
I just pushed a fix for appending across different projects, which I call out in the test instructions above.

akofink:
Nice catch! It was very reproducible 😅

I've added an early return for the nil/empty session name, which is expected outside a tmux session..

akofink:
> Does this run `pre_window` on those windows that are appended? If so, is there maybe a way to tell it not to run `pre_window` when appending?

This change as written runs `pre_window` for appended windows in the existing session.

Perhaps it'd be good to implement a `--no-pre-window` option to turn it off that would work with or without `--append`.

hschne-ml:
> I've added an early return for the nil/empty session name, which is expected outside a tmux session..

Confirmed OK, works now :+1:

@akofink Thank you so much for getting back to this PR so quickly! Let's ship it! :grin:  :ship:

akofink:
Thanks again for your help, @hschne! I really appreciate it.

After merge, it won't be available in tmuxinator until we draft a release, which I will do later this week.

PR #913, opened 2024-04-23
Fix pane title options

augustelalande:
I did not do a very good job testing in #892, because the current implementation of `enable_pane_titles`, `pane_title_position`, and `pane_title_format` only sets the desired setting for the first window. I think I may have had a previous global configuration active which threw off my results.

This PR fixes those options to behave as expected, i.e., setting the options across the project.

As it turns out `tmux set pane-border-status top` only sets the option for a single window, i.e., the active one. For it to be set across a session without using the global `-g` flag, the command must target each window individually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- Fix pane title configuration options only applying to first window
 
 ## 3.2.0
 ### Third-party Dependencies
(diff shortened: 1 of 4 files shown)


akofink (approved):
Works well and fixes the bug - thank you! 🥇

Would you mind adding a line in the changelog? When I tested the original PR, I only used 1 window 😿

augustelalande:
Done. Sorry keep forgetting.

augustelalande:
@akofink Can you make a release for this fix?

Issue #909, opened 2024-04-10
Unable to successfully run tmuxinator on macOS Sonoma 14.4.1


s5b:
**Describe the bug**
A couple of weeks ago I was able to run a number of projects with `tmuxinator` on my Mac Book Pro without problem, as I have been doing for a number of years now. Today, however, running `tmuxinator` as I've always done simply opens any configured panes and runs the first command associated with each _pane_ and then stops.

**To Reproduce**
I appear to be unable run any of my existing projects, or any _new_ projects with `tmuxinator`. The problem is that all expected _panes_ open open start (good), but all commands are simply echoed at the top of each pane, and then only the first command appears to be executed. All subsequent commands are ignored.

**Expected behavior**
I expect each _pane_ to open, and for the commands _not_ to be echoed as a single block at the top of the _pane_, and for each command to be executed in sequence.

**Environment**
 - tmuxinator Version: `tmuxinator 3.1.2`
- tmux Version: `tmux 3.4`
 - Ruby Version: `ruby 2.5.5p157 (2019-03-15 revision 67260) [-darwin21]`
 - OS: `macOS Sonoma 14.4.1`

**Additional information**
Running this project will simply list the files in each `~/tmp/tmuxbuginator` directory, but will _not_ `cat` the contents of the two files:

```
name: tp-experiment
root: ~/tmp/tmuxbuginator

# Optional tmux socket
# socket_name: foo

# Note that the pre and post options have been deprecated and will be replaced by
# project hooks.

# Project hooks
# Runs on project start, always
# on_project_start: command
# Run on project start, the first time
# on_project_first_start: command
# Run on project start, after the first time
# on_project_restart: command
# Run on project exit ( detaching from tmux session )
# on_project_exit: command
# on_project_exit: tmux -CC attach
# Run on project stop
# on_project_stop: command

# Runs in each window and pane before window/pane specific commands. Useful for setting up interpreter versions.
# pre_window: rbenv shell 2.0.0-p247

# Pass command line options to tmux. Useful for specifying a different tmux.conf.
# tmux_options: -f ~/.tmux.mac.conf
# tmux_options: -f ~/.tmux.conf

# Change the command to call tmux.  This can be used by derivatives/wrappers like byobu.
# tmux_command: byobu

# Specifies (by name or index) which window will be selected on project startup. If not set, the first window is used.
# startup_window: editor

# Specifies (by index) which pane of the specified window will be selected on project startup. If not set, the first pane is used.
[...]

PR #914, opened 2024-04-24
Prioritize newer syntax if available

augustelalande:
Due to my investigation into #909 I have become keenly aware of the printouts tmuxinator may produce. One such printout is the `cd <root>` which gets sent to the first window in the project. What's more, this is not necessary with the quote-unquote modern (tmux >1.9 lol) syntax, where a starting directory can be provided directly to the `new-session` command.

This PR will take advantage of this syntax if available to reduce the excessive printouts.

N.B.: @akofink I'm starting to see why you wanted to drop the earlier versions of tmux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 - Fix pane title configuration options only applying to first window
+- Use `tmux -c` if available to reduce spam
 
 ## 3.2.0
 ### Third-party Dependencies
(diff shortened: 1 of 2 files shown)


akofink (approved):

akofink:
Hey @augustelalande - thanks again! I really appreciate your recent contributions. This makes sense to me. It just needs a changelog entry. 🙏

augustelalande:
Done

PR #917, opened 2024-06-04
fix: Session current path

### LINKED ISSUE #916: `project.root` property is "overridden" by `windows.first.root`
**Describe the bug**
When using version 3.2.1, if `windows.first.root` is defined, it overrides the `project.root` configuration - starting new windows with `tmux` will start the shell at `windows.first.root` instead of `project.root`.

I believe this is caused by the recently added `-c` in `tmux new-session` (see the output of `tmuxinator debug` below) in version 3.2.1.
I have compared the behaviour to 3.2.0, and in 3.2.0 it works as I expect (the culprit is possibly #914).
In 3.2.0 the `windows.first.root` used `cd`, which only affected the shell itself. Whereas `-c` changes the tmux session.

I have found a workaround for myself: detach the tmux session and reattach to it using `-c <project.root>` arg.
Integrated this with an `on_project_exit` hook, e.g. `on_project_exit: tmux attach-session -t bug -c ~/parent`

**To Reproduce**
On version 3.2.1, use the following config:
```yaml
name: bug
root: ~/parent

windows:
  - dir2:
      root: ~/parent/dir2
  - dir1:
      root: ~/parent/dir1
```

Start the session:
```bash
mkdir -p ~/parent/dir{1,2}
tmuxinator start sample
```
Create a new window in tmux. The current working directory is set to `~/parent/dir2` instead of `~/parent`.

**Expected behavior**
When creating new windows in the given session, I expect the new window to be created at the path `~/parent`.
This however does not happen; the new window is created at `~/parent/dir2`.

**Environment**
 - tmuxinator Version: tmuxinator 3.2.1
- tmux Version: tmux 3.4
 - Ruby Version: ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [x86_64-linux]
 - OS: Arch Linux

**Additional information**
Output of `tmuxinator debug bug` using 3.2.1:
```sh
#!/bin/zsh

# Clear rbenv variables before starting tmux
unset RBENV_VERSION
unset RBENV_DIR

tmux start-server;

cd /home/user/parent

# Run on_project_start command.



  # Run pre command.


  # Run on_project_first_start command.


  # Create the session and the first window. Manually switch to root
  # directory if required to support tmux < 1.9
  TMUX= tmux new-session -d -s bug -n dir2 -c /home/user/parent/dir2



  # Create other windows.
  tmux new-window -c /home/user/parent/dir1 -t bug:2 -n dir1


  # Window "dir2"




  # Window "dir1"




  tmux select-window -t bug:1
  tmux select-pane -t bug:1.1

  if [ -z "$TMUX" ]; then
    tmux -u attach-session -t bug
  else
    tmux -u switch-client -t bug
  fi



# Run on_project_exit command.
```

Output of `tmuxinator debug sample` using 3.2.0:
```sh
#!/bin/zsh

# Clear rbenv variables before starting tmux
unset RBENV_VERSION
unset RBENV_DIR

tmux start-server;

cd /home/user/parent

# Run on_project_start command.



  # Run pre command.


  # Run on_project_first_start command.


  # Create the session and the first window. Manually switch to root
  # directory if required to support tmux < 1.9
  TMUX= tmux new-session -d -s bug -n dir2
  tmux send-keys -t bug:1 cd\ /home/user/parent/dir2 C-m



  # Create other windows.
  tmux new-window -c /home/user/parent/dir1 -t bug:2 -n dir1


  # Window "dir2"


  # Window "dir1"


  tmux select-window -t bug:1
  tmux select-pane -t bug:1.1

  if [ -z "$TMUX" ]; then
    tmux -u attach-session -t bug
  else
    tmux -u switch-client -t bug
  fi



# Run on_project_exit command.
```

akofink:
## Problems being solved

We [attempted to leverage](https://github.com/tmuxinator/tmuxinator/pull/914) `tmux new-session [-c start-directory]` for tmux >= 1.9 instead of relying on `cd <start-directory>` before the `new-session` command, which by default uses the current working directory.

The tmux manpage for `new-session` allows configuring the first window name, but `-c` still applies to the session:

```
new-session [-AdDEP] [-c start-directory] [-F format] [-n window-name] [-s session-name] [-t group-name] [-x width] [-y height] [shell-command]
             (alias: new)
       Create a new session with name session-name.

       The new session is attached to the current terminal unless -d is given.  window-name and shell-command are the name
       of and shell command to execute in the initial window.  If -d is used, -x and -y specify the size of the initial
       window.
       ...
```

The PR that introduced `-c` for `new-session` used the first window's root, if it was defined, instead of the project's root.

Fixes #916 

## How the problems are being solved

Remove the `-c` option in `new-session` entirely to use the current working directory, set to the project root vs. the first window's root. This allows us to remove special handling for tmux < 1.9, which is when the option was added to `new-session` (it exists on `new-window` since before tmux 1.0).

Run the `new-window` command for all windows, including the first window, with `-c`, leveraging option `-k`:

```
if the target already exists an error is shown, unless the -k flag is used, in which case it is destroyed.
...
-c specifies the working directory in which the new window is created.
```

## How to test it

Use this config, copied from #916:

```yaml
name: 916
root: /tmp

windows:
  - dir2:
      root: /tmp/dir2
  - dir1:
      root: /tmp/dir1
```

`mux debug 916` output shows correct `-c` paths:

```sh
  ...
  cd /tmp
  ...
  TMUX= tmux new-session -d -s 916 -n a

  # Create windows.
  tmux new-window -c /tmp/dir2 -k -t 916:1 -n dir2
  tmux new-window -c /tmp/dir1 -k -t 916:2 -n dir1
  ...
```

### DIFF
--- a/lib/tmuxinator/assets/template.erb
+++ b/lib/tmuxinator/assets/template.erb
@@ -19,18 +19,7 @@ cd <%= root || "." %>
   # Run on_project_first_start command.
   <%= hook_on_project_first_start %>
 
-  # Create the session and the first window. Manually switch to root
-  # directory if required to support tmux < 1.9
-  <% if windows.first.root? %>
-    <% if Tmuxinator::Config.version < 1.9 %>
   TMUX= <%= tmux_new_session_command %>
-  <%= windows.first.tmux_window_command_prefix %> <%= "cd #{windows.first.root}".shellescape %> C-m
-    <%- else -%>
-  TMUX= <%= tmux_new_session_command %> -c <%= windows.first.root.shellescape %>
-    <% end %>
-  <%- else -%>
-  TMUX= <%= tmux_new_session_command %>
-  <% end %>
 
   <% if Tmuxinator::Config.version < 1.7 %>
   # Set the default path for versions prior to 1.7
@@ -48,8 +37,8 @@ cd <%= root || "." %>
     <% end %>
   <% end %>
 
-  # Create other windows.
-  <% windows.drop(1).each do |window| %>
+  # Create windows.
+  <% windows.each do |window| %>
   <%= window.tmux_new_window_command %>
   <% end %>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## Unreleased
+### Fixes
+- Session path is project root, not first window root
 
 ## 3.2.1
 ### Enhancements
--- a/lib/tmuxinator/window.rb
+++ b/lib/tmuxinator/window.rb
@@ -115,7 +115,7 @@ module Tmuxinator
 
     def tmux_new_window_command
       path = root? ? "#{Tmuxinator::Config.default_path_option} #{root}" : nil
-      "#{project.tmux} new-window #{path} -t #{tmux_window_target} #{tmux_window_name_option}"
+      "#{project.tmux} new-window #{path} -k -t #{tmux_window_target} #{tmux_window_name_option}"
     end
 
     def tmux_tiled_layout_command
--- a/spec/lib/tmuxinator/window_spec.rb
+++ b/spec/lib/tmuxinator/window_spec.rb
@@ -354,7 +354,7 @@ describe Tmuxinator::Window do
 
     let(:window_part) { "new-window" }
     let(:name_part) { window.tmux_window_name_option }
-    let(:target_part) { "-t #{window.tmux_window_target}" }
+    let(:target_part) { "-k -t #{window.tmux_window_target}" }
     let(:path_part) { "#{path_option} #{project.root}" }
 
     let(:path_option) { "-c" }


lib/tmuxinator/assets/template.erb
```diff
@@ -19,18 +19,7 @@ cd <%= root || "." %>
...
-  <% if windows.first.root? %>
-    <% if Tmuxinator::Config.version < 1.9 %>
   TMUX= <%= tmux_new_session_command %>
-  <%= windows.first.tmux_window_command_prefix %> <%= "cd #{windows.first.root}".shellescape %> C-m
-    <%- else -%>
-  TMUX= <%= tmux_new_session_command %> -c <%= windows.first.root.shellescape %>
-    <% end %>
-  <%- else -%>
-  TMUX= <%= tmux_new_session_command %>
-  <% end %>
```
akofink, line 33: All this was trying to set up the session path without breaking old tmux versions. The logic is mostly the same, but it's moved into `project.rb`.

lib/tmuxinator/assets/template.erb
```diff
@@ -48,8 +37,8 @@ cd <%= root || "." %>
     <% end %>
   <% end %>
 
-  # Create other windows.
-  <% windows.drop(1).each do |window| %>
+  # Create windows.
+  <% windows.each do |window| %>
```
akofink, line 41: Using `new-window` with the `-k` option allows us to treat the first window, which is created when we create the session, like all the other windows.

augustelalande: Clean

lib/tmuxinator/project.rb
```diff
@@ -328,9 +328,15 @@ def show_tmux_options
...
+    def tmux_new_session_path
+      if root? && Tmuxinator::Config.version >= 1.9
+        "#{Tmuxinator::Config.default_path_option} #{root}"
+      end
+    end
+
     def tmux_new_session_command
       window = windows.first.tmux_window_name_option
-      "#{tmux} new-session -d -s #{name} #{window}"
+      "#{tmux} new-session #{tmux_new_session_path} -d -s #{name} #{window}"
```
augustelalande, line 339: If I understand correctly, this whole section doesn't really matter because of this
https://github.com/tmuxinator/tmuxinator/blob/0f7e9ffa93c508c45bb0bfd7537a7b5b675e0048/lib/tmuxinator/assets/template.erb#L9

akofink: It's true that the `-c` option's default is the current working directory.

It doesn't hurt to both `cd` and set `-c` here, even when using `mux start --no-attach [PROJECT]`. I tried this, and the CWD was not changed, despite the `cd` command being in the shell script output with `mux debug --no-attach [PROJECT]`.

Still, I guess we can completely remove `tmux_new_session_path` 🤔

akofink: I tried removing the `-c` entirely to rely on the current working directory, and the session path still seems to get set properly. I'll update the PR.

augustelalande: Should we consider dropping support for tmux < 1.9, and then getting rid of the `cd` instead?

akofink: I'm not sure what removing the `cd` and relying on `-c` gets us. Keeping it allows us to continue supporting the older tmux versions, which is a plus IMO.

augustelalande: fair

Issue #857, opened 2022-05-26
Fish completion is broken


legeana:
**Describe the bug**
tmuxinator/completion/tmuxinator.fish doesn't correctly set __fish_tmuxinator_program_cmd in `set __fish_tmuxinator_program_cmd (commandline -o)[1]`. In my shell, tested in empty fish environment as well, this variable is empty after I try reproducing this problem.

```
$ set -q __fish_tmuxinator_program_cmd
> returns 0, the variable is set
$ count __fish_tmuxinator_program_cmd
1
```

Copying `tmuxinator.fish` into `~/.config/fish/completions/tmuxinator.fish` and setting `__fish_tmuxinator_program_cmd` to `tmuxinator` instead of `(commandline -o)[1]` mitigates the problem. As far as I understand `(commandline -o)[1]` returns empty string.

My guess that setting it to `(commandline -o)[1]` was intended to make it work with both `mux` and `tmuxinator` commands, but I don't think that's necessary because `mux` expands into `tmuxinator` anyway. Or if needed maybe a for-loop and explicitly setting both might be a better solution. I am not quite sure.

**To Reproduce**
Typing `tmuxinator ` (notice space at the end) triggers the following

```
~> tmuxinator complete: too many arguments

/usr/share/fish/vendor_completions.d/tmuxinator.fish (line 17):
complete -f -c $__fish_tmuxinator_program_cmd                                                -a '(__fish_tmuxinator_program completions start)'
^
from sourcing file /usr/share/fish/vendor_completions.d/tmuxinator.fish
in command substitution

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/tmuxinator.fish (line 18):
complete -f -c $__fish_tmuxinator_program_cmd -n '__fish_use_subcommand'                  -x -a "(__fish_tmuxinator_program commands)"
^
from sourcing file /usr/share/fish/vendor_completions.d/tmuxinator.fish
in command substitution

(Type 'help complete' for related documentation)
complete: too many arguments

/usr/share/fish/vendor_completions.d/tmuxinator.fish (line 19):
complete -f -c $__fish_tmuxinator_program_cmd -n '__fish_tmuxinator_using_command start'     -a "(__fish_tmuxinator_program completions start)"
^
from sourcing file /usr/share/fish/vendor_completions.d/tmuxinator.fish
in command substitution

(Type 'help complete' for related documentation)
complete: too many arguments

[...]

PR #956, opened 2025-06-16
fix: Fix `fish-shell` completions (857)

Drew-Daniels:
Refactors `fish-shell` completion so that it works as expected.

### Metadata

Fixes https://github.com/tmuxinator/tmuxinator/issues/857

### Problem / Motivation

Fixing `fish-shell` completions. In my case, I was noticing that completions for `tmuxinator` did not work at all, either when typing `tmuxinator<TAB>` or when using my `mux` alias set in my `~/.config/fish/config.fish` file, via `alias mux="tmuxinator"`.

### Solution

- [x] CHANGELOG entry is added for code changes

Removes some problematic logic around getting a reference to the un-aliased/aliased `tmuxinator` command a-la `set __fish_tmuxinator_program_cmd (commandline -o)[1]`. 

If `fish` users are using an alias like `mux`, `fish` should already be able to map this to the underlying `tmuxinator` command: https://fishshell.com/docs/current/cmds/alias.html#example, so this logic is unnecessary and was the cause that fish shell completions were not working. By replacing references to this with `tmuxinator`, the shell completion now works

#### QOL Updates
Unnecessary, but I've made a couple additional updates for QOL purposes:
- Using the long option names for `complete` for better readability.
- Adds completions for the `edit` and `debug` commands. They use the same completion logic as `tmuxinator completions start`, so it's sort of a hack, but since we should be able to edit/debug the same projects, I just used the same underlying command to get the available projects.

### Testing

| Before | After |
| -------------- | --------------- |
| mux_commands_before | mux_commands|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Document new from session feature in the README
 ## Fixes
 - Properly pass args with equals (=) in their values
+- Fix `fish-shell` completion
 
 ### Features
 - Add support for tmuxinator stop-all
--- a/completion/tmuxinator.fish
+++ b/completion/tmuxinator.fish
@@ -1,22 +1,17 @@
 function __fish_tmuxinator_using_command
-  set cmd (commandline -opc)
-  if [ (count $cmd) -gt 1 ]
-    if [ $argv[1] = $cmd[2] ]
-      return 0
+    set cmd (commandline -opc)
+    if [ (count $cmd) -gt 1 ]
+        if [ $argv[1] = $cmd[2] ]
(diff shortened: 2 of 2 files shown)


akofink (approved):
Thanks @Drew-Daniels! The change makes sense to me and fixes the completions. I also tested with the `mux` alias in fish, and it still seems to work fine. 🎉

akofink:
Not clear if it was ever actually needed, or ever actually worked, but it was introduced here, for reference: https://github.com/tmuxinator/tmuxinator/pull/280

PR #959, opened 2025-07-24
Update Thor dependency and everything else too :see_no_evil: 

### LINKED ISSUE #958: pinned version of `thor` has an open security advisory
**Describe the bug**
[Here](https://github.com/tmuxinator/tmuxinator/blob/fe37bf00469f7fb6fcdfa0a0ee7db2638877816f/tmuxinator.gemspec#L44) `thor` is added as a dependency to this project, with the version `~> 1.3.0`. This means version `1.4.0` is not allowed. All versions of `thor` below `1.4.0` are affected by a vulnerability noted [here](https://github.com/advisories/GHSA-mqcp-p2hv-vw6x). I would like to use `tmuxinator` and use a secure version of `thor`.

hschne:
### Metadata

Fixes #958 

### Solution

This PR updates the linked Thor dependency, and then some more. Couldn't help myself
- Updated Thor to fix security advisory
- Updated Ruby to 2.7. Not sure if we should update to 3.1, since we're not even running tests for older versions? :thinking: 
- Updated the runtime dependencies. XDG update had some breaking changes, which required updates to `lib/tmuxinator/config.rb` and related test changes.
- Update dev dependencies. `amazing_print` replaces `awesome_print`. The Rubocop update required numerous style fixes, which accounts for most of the changes. 
- Removed Gemfile.lock from gitignore. Checking in the lockfile is generally recommended, not sure why it wasn't in the first place. 

I know it's a lot, but better to ask for forgiveness than for permission. Happy to trim this down to make the PR more manageable, just let me know.


CHANGELOG.md
```diff
@@ -2,13 +2,22 @@
...
 - Fix `fish-shell` completion
 
 ### Features
 - Add support for tmuxinator stop-all
 
+### Features
+- Add support for tmuxinator stop-all
+
+### Features
+- Add support for tmuxinator stop-all
```
akofink, line 19: This seems like it was added multiple times on accident, maybe just a bad merge resolution?

Gemfile.lock
```diff
@@ -0,0 +1,121 @@
+PATH
```
akofink, line 1: I believe it's best practice to check this in for applications but not rubygems that need to work in a variety of different environments/ruby versions.

akofink (approved):
Thanks @hschne! The changes look good to me.

Issue #979, opened 2026-04-08
Project config inheritance?


mockdeep:
**Is your feature request related to a problem? Please describe.**

I'm not seeing a way to inherit from another config.

**Describe the solution you'd like**

I would love it if there was a way to create a high level config that can be inherited for projects. For example, I might create a `rails.yml` file, and for Rails projects I would inherit from that and add paths and variations in `my-rails-project.yml`.

**Describe alternatives you've considered**

Maybe using ERB to read in a generic config? That might get a bit more complicated if it comes to merging nested keys, though.

**Additional context**

Love this project, thanks!

akofink:
An interesting idea - it reminds me of what docker compose does for its overrides, which as far as I know is a bespoke implementation with fairly complex rules. It is not a trivial problem to solve.

That said, one way tmuxinator projects are able to handle multiple different configurations is [by using `@args`/`@settings` within the template](https://github.com/tmuxinator/tmuxinator#erb). So for the rails example:

```yaml
name: rails-<%= @args[0] %>
root: ~/dev/repos/rails/<%= @args[0] %>

windows:
  ...
```

and then `tmuxinator start rails my-project`

---

I do not recommend doing this and no guarantees if it one day breaks, but the way ERB templates are rendered gives you basically infinite flexibility to do whatever you want. So a project `test2` that inherits all the windows of `test1`:

```yaml
<%= render_template("#{path[/.*\/(?=[^\/]+\.yml)/]}test1.yml", binding) %>
  - morewindows:
      - bundle exec rubocop
name: test2
```

The last top level key is kept (i.e. `name` has to go after the windows). This results in a combined project config like:

```yaml
name: test2
root: ~/dev/

windows:
  - zsh:
  - vim: vim
  - morewindows:
      - bundle exec rubocop
```

🙈 don't do it like this. Also, my editor hates me trying to treat this like any kind of legit yaml.

ethagnawl:
There could be a clever way to achieve this by introducing a top-level "partial" or "base" config option and using YAML anchors and aliases. (Perhaps plural because that would be an obvious evolution.)

- The partial path lookup could use the same/similar logic that the project config does or be explicitly passed
- The contents would get loaded from the file -- most likely as a well-formed anchor to keep it simple
- There should probably be a validation pass. (Maybe the existing logic could be re-used? It's been a while since I've thought about any of this.)
[...]

PR #987, opened 2026-05-15
Introduce "partials" to support config inheritance 

ethagnawl:
### Metadata

This is a first attempt at addressing the "project config inheritance" feature requested in #979 

### Problem / Motivation

This change set introduces a new top-level config property called "partials". It allows the user to specify a list of partial or base (I'm not satisfied with either name yet) config files that their project config will extend.

This is still WIP but the current strategy collapses all top-level config options and the last specified in the list wins. As far as windows go, the current iteration accrues them as it processes the partials and eventually the project config and doesn't attempt to do anything clever about merging windows with the same name, deduping commands, etc. There are a few ways this could be made better but this initial pass is simple and unsurprising. We could potentially make the window treatment configurable but I think that's best left for a future PR.

Errors resulting from the parsing and merging of partials bubble up to the exsting `#load` exception handler and are handled in the same way: tmuxinator exits with an error message. We _could_ handle the errors and continue with just the project but, again, I think simple and obvious is the preferred approach.

### Solution

In practice, the feature looks like:

```
# foo.yaml
partials:
  - "/home/peter/.config/tmuxinator/partial-1.yaml"
  - "/home/peter/.config/tmuxinator/partial-2.yaml"
name: foo
attach: true
windows:
  - editor:
      layout: main-vertical
      panes:
        - editor:
          - vim

# partial-1.yaml
pre:
  - echo "partial is working"
  
# partial-2.yaml 
socket_name: qux
windows:
  - editor:
      layout: main-vertical
      panes:
        - editor:
          - emacs
```

The resulting tmuxinator commands are as follows:

```
#!/bin/bash


  # Clear rbenv variables before starting tmux
  unset RBENV_VERSION
  unset RBENV_DIR

  tmux -L qux start-server;

cd .

# Run on_project_start command.



  # Run pre command.
  

  # Run on_project_first_start command.
  

  tmux -L qux new-session -d -s foo -n editor



  # Create windows.
  tmux -L qux new-window  -k -t foo:0 -n editor
  tmux -L qux new-window  -k -t foo:1 -n editor


  # Window "editor"


  tmux -L qux select-pane -t foo:0.0 -T editor
  tmux -L qux send-keys -t foo:0.0 emacs C-m

  tmux -L qux select-layout -t foo:0 tiled

  tmux -L qux select-layout -t foo:0 main-vertical
  tmux -L qux select-pane -t foo:0.0


  # Window "editor"


  tmux -L qux select-pane -t foo:1.0 -T editor
  tmux -L qux send-keys -t foo:1.0 vim C-m

  tmux -L qux select-layout -t foo:1 tiled

  tmux -L qux select-layout -t foo:1 main-vertical
  tmux -L qux select-pane -t foo:1.0


  tmux -L qux select-window -t foo:0
  tmux -L qux select-pane -t foo:0.0

  if [ -z "$TMUX" ]; then
    tmux -L qux -u attach-session -t foo
  else
    tmux -L qux -u switch-client -t foo
  fi



# Run on_project_exit command.
```

### TODO
- [x] Use project config lookup for partials so full paths aren't required
- [ ] Decide if the current window treatment is satisfactory
- [ ] Test degenerate cases:
  - [ ] Malformed partial
  - [ ] Missing partial
- [ ] Change "partials" to ... something else?
- [ ] Anything else?

Any thoughts about this feature at either the micro or macro levels, @tmuxinator/tmuxinator?


lib/tmuxinator/project.rb
```diff
@@ -48,15 +48,45 @@ def load(path, options = {})
...
         new(yaml, options)
       end
 
+      def load_partials(project_yaml)
+        partials = project_yaml.fetch("partials", []) || []
+        partials.inject({}) do |memo, partial|
+          partial_content = render_template(partial, binding)
+          partial_yaml_ = YAML.safe_load(partial_content, aliases: true)
+          memo.merge(partial_yaml_)
+        end
```
akofink: Multiple partials do not actually accumulate windows from earlier partials. `load_partials` merges each partial with plain `Hash#merge`, so if `partial-1.yml` and `partial-2.yml` both define `windows`, the second partial overwrites the first before `load_and_merge_partials` ever gets a chance to concatenate arrays with the project config.

This contradicts the PR behavior described as accruing windows while processing partials and means users lose windows depending on partial ordering. The merge block that handles array composition below ideally needs to be applied while folding partials too, or specifically for `windows` across all partials plus the project.

lib/tmuxinator/project.rb
```diff
@@ -48,15 +48,45 @@ def load(path, options = {})
...
+        # The merge block currently exists to ensure that windows arrays are
+        # composed instead of clobbered. This first version is naive and
+        # doesn't attempt to merge windows contents and accumulates them. The
+        # result is that you may wind up with multiple windows with the same
+        # name and, perhaps, the same content. This is better than clobbering
+        # windows as partials are processed but the ideal is probably to merge
+        # and dedupe like-windows. We'll see ...
+        partial_yamls.merge(project_yaml) do |_key, partial_val, project_val|
+          if partial_val.is_a?(Array) && project_val.is_a?(Array)
+            partial_val + project_val
```
akofink: The array merge applies to every top-level array, not just `windows`. That changes override semantics for options such as `pre`, `post`, `pre_window`, and project hooks: a project that defines its own `pre` array will unexpectedly run the partial's `pre` commands first instead of replacing the partial value. The PR description says top-level options should be last-one-wins while only windows accrue, so this should probably check `_key == "windows"` before concatenating.

akofink (commented):
Thanks for getting this functional PoC written up!

Similar to my comments on the related issue/feature request, I think this feature is useful but should be weighed against tmuxinator's existing args/options mechanism (`mux <project> <arg> <option>=<value>`).

Partials and runtime args solve related extensibility problems from different directions: args are good for parameterizing a single project at invocation time, while partials are better for sharing reusable baseline config across projects.

That distinction may justify having both, but it would be worth keeping the merge contract very small and explicit so the feature remains as easy to reason about as these helper methods currently are.

lib/tmuxinator/project.rb
ethagnawl: Thanks. This is also a great point. I'd forgotten there were other top-level options which could also be arrays. (I wish there was a typed project schema to reference ... ). The current, permissive behavior actually results in some weird edge cases:
Screenshot from 2026-06-01 13-16-04
Screenshot from 2026-06-01 13-14-50

I might open a separate issue to track the above.

Anyways, in those cases where arrays are valid for top-level options, I _think_ it would probably make sense to have these accumulate, too, and only use the last-one-wins behavior in cases where accumulation isn't possible.

lib/tmuxinator/project.rb
ethagnawl: You're absolutely right. Thanks for catching that.

ethagnawl:
I'm going to close this for now. I may pick it back up at some point if I find time or if anyone else starts actively asking for this feature.

PR #927, opened 2024-09-27
Add tmux 3.5 to the list of supported tmux versions

farzadmf:
### Metadata

N/A

### Problem / Motivation

`tmux` was recently updated to version `3.5`, but tmuxinator shows it as unsupported.

### Solution

I'm updating `tmux_versions.rb` file to add the version.

### Testing

Didn't test; let me know if there are tests to be run/added/updated.

I have `tmux 3.5` installed, and I get this message when starting `tmuxinator`:
image

PR #997, opened 2026-06-29
chore: add support for tmux 3.7

iyuuya:
### Metadata

https://github.com/tmux/tmux/releases/tag/3.7

### Problem / Motivation

tmux 3.7 released.

### Solution

Add tmux 3.7 to supported version list.

### Testing

Updated config_spec.rb


farzadmf:
@iyuuya Take a look at PR #927 ; there's an additional file for CI that needs to be updated as well

akofink (approved):

iyuuya:
Thanks, @farzadmf! That CI-file update was the old approach from before #981. #981 reworked CI to use snapshot-based interface tests instead of a tmux version matrix, so new tmux versions no longer need a CI file change — just SUPPORTED_TMUX_VERSIONS + CHANGELOG + spec update, as #989 did for 3.6b. This PR follows that same pattern.

gstokkink-bex:
@akofink can we get a release for this? thanks 😄

farzadmf:
> Thanks, @farzadmf! That CI-file update was the old approach from before #981. #981 reworked CI to use snapshot-based interface tests instead of a tmux version matrix, so new tmux versions no longer need a CI file change — just SUPPORTED_TMUX_VERSIONS + CHANGELOG + spec update, as #989 did for 3.6b. This PR follows that same pattern.

Oh, good to know; thanks for letting me know @iyuuya

